PR #11, opened 2020-09-23
Checks for git access token if status tracker enabled

gaganhegde:
**What type of PR is this?**
> /kind cleanup

This PR makes sure that the git token is asked for if the commit status tracker is enabled.

https://issues.redhat.com/browse/GITOPS-347


pkg/cmd/bootstrap.go
```diff
@@ -120,6 +120,9 @@ func nonInteractiveMode(io *BootstrapParameters, client *utility.Client) error {
 			return fmt.Errorf("The mandatory flag %q has not been set", value)
 		}
 	}
+	if io.CommitStatusTracker && io.GitHostAccessToken == "" {
+		return fmt.Errorf("Kindly input the git-host-access-token if commit-status-tracker set to true")
```
bigkevmcd, line 124: This doesn't need to be `Errorf` there's no format strings in the string?

Also, let's maybe tweak that to `--git-host-access-token is required when commit-status-tracker is enabled` ?

bigkevmcd (dismissed):
Looks good, a minor fix and should be good to go.

pkg/pipelines/statustracker/deployment.go
```diff
@@ -108,6 +112,10 @@ func makeEnvironment(repoURL, driver string) []corev1.EnvVar {
 			Name:  "OPERATOR_NAME",
 			Value: operatorName,
 		},
+		{
+			Name:  "STATUS_TRACKER_SECRET",
+			Value: "git-host-access-token",
```
wtam2018, line 117: Use this const CommitStatusTrackerSecret?

wtam2018: Anywhere using this "git-host-access-token" string should use the constant?  If the secret constant shall be used not limited to commit status tracker, consider renaming the const.

wtam2018 (changes requested):
I've added some questions to be addressed.

pkg/cmd/bootstrap.go
```diff
@@ -13,8 +13,10 @@ import (
 	"github.com/redhat-developer/kam/pkg/cmd/utility"
 	"github.com/redhat-developer/kam/pkg/pipelines"
 	"github.com/redhat-developer/kam/pkg/pipelines/ioutils"
+	"github.com/redhat-developer/kam/pkg/pipelines/statustracker"
 	"github.com/spf13/cobra"
 
+	err "github.com/pkg/errors"
```
wtam2018, line 19: Not sure if we want to use "err" as an aliase because we have "err" as  variable name all over the place.

bigkevmcd: It's also not clear why this is importing `pkg/errors` at all?

`pkg/errors` is deprecated as most of the functionality was folded into Go.

wtam2018: +1

```
import "errors"

err := errors.New(" ...")
```

bigkevmcd (approved):

wtam2018 (commented):
Add a note about moving token check flag to `Validate()`

pkg/cmd/bootstrap.go
```diff
@@ -121,7 +123,8 @@ func nonInteractiveMode(io *BootstrapParameters, client *utility.Client) error {
 	if err := checkMandatoryFlags(mandatoryFlags); err != nil {
 		return err
 	}
-	if err := checkBootstrapDependencies(io, client, log.NewStatus(os.Stdout)); err != nil {
+	err := checkBootstrapDependencies(io, client, log.NewStatus(os.Stdout))
+	if err != nil {
 		return err
```
wtam2018, line 128: Please revert this change.

pkg/cmd/bootstrap.go
```diff
@@ -298,12 +302,12 @@ func NewCmdBootstrap(name, fullName string) *cobra.Command {
...
 	bootstrapCmd.Flags().BoolVar(&o.Overwrite, "overwrite", false, "Overwrites previously existing GitOps configuration (if any)")
 	bootstrapCmd.Flags().StringVar(&o.ServiceRepoURL, "service-repo-url", "", "Provide the URL for your Service repository e.g. https://github.com/organisation/service.git")
 	bootstrapCmd.Flags().StringVar(&o.ServiceWebhookSecret, "service-webhook-secret", "", "Provide a secret that we can use to authenticate incoming hooks from your Git hosting service for the Service repository. (if not provided, it will be auto-generated)")
 	bootstrapCmd.Flags().StringVar(&o.PrivateRepoDriver, "private-repo-driver", "", "If your Git repositories are on a custom domain, please indicate which driver to use github or gitlab")
-	bootstrapCmd.Flags().BoolVar(&o.CommitStatusTracker, "commit-status-tracker", true, "Enable or disable the commit-status-tracker which reports the success/failure of your pipelineruns to GitHub/GitLab")
+	bootstrapCmd.Flags().BoolVar(&o.CommitStatusTracker, "commit-status-tracker", false, "Enable or disable the commit-status-tracker which reports the success/failure of your pipelineruns to GitHub/GitLab")
```
wtam2018, line 310: Are you intentionally changing the default to false?

pkg/cmd/bootstrap_test.go
```diff
@@ -103,10 +103,10 @@ func TestValidateCommitStatusTracker(t *testing.T) {
 		gitAccessToken      string
 		wantErr             string
 	}{
-		{"statusTracker true/ GitAccessToken absent", "username1/testRepo1", true, "", "--git-host-access-token is required if commit-status-tracker is enabled"},
+		{"statusTracker true/ GitAccessToken present", "username1/testRepo1", true, "", "--git-host-access-token is required if commit-status-tracker is enabled"},
 		{"statusTracker true/ GitAccessToken absent", "username2/testRepo2", true, "abc123", ""},
 		{"statusTracker false/ GitAccessToken absent", "username3/testRepo3", false, "abc123", ""},
-		{"statusTracker false/ GitAccessToken present", "username3/testRepo3", false, "abc123", ""},
+		{"statusTracker false/ GitAccessToken absent", "username3/testRepo3", false, "abc123", ""},
 	}
```
wtam2018, line 110: Still wrong.  @gaganhegde

wtam2018 (approved):

pkg/cmd/bootstrap_test.go
gaganhegde: I think the change was addressed William! my bad forgot to save.

PR #8, opened 2020-09-23
add setup devconsole page

wtam2018:
**What type of PR is this?**
> Uncomment only one ` /kind` line, and delete the rest.
> For example, `> /kind bug` would simply become: `/kind bug`

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind failing-test
/kind feature
> /kind flake
> /kind code-refactoring

**What does this PR do / why we need it**:
Add setup devconsole document page
**Have you updated the necessary documentation?**

* [x] Documentation update is required by this PR.
* [x] Documentation has been updated.

**Which issue(s) this PR fixes**:

Fixes #?
N/A
**How to test changes / Special notes to the reviewer**:
N/A

PR #24, opened 2020-09-24
Expand file path before writing resources

chetan-rns:
**What type of PR is this?**
> /kind bug

* The interactive prompt can now accept relative path and tilde(`~`)
* Add .git suffix before checking dependencies
* Move the completion message to Run(), to keep the UX similar in both the modes

Fixes #?

### DIFF
--- a/pkg/pipelines/yaml/resources_test.go
+++ b/pkg/pipelines/yaml/resources_test.go
@@ -1,0 +1,56 @@
+package yaml
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/redhat-developer/kam/pkg/pipelines/namespaces"
+	res "github.com/redhat-developer/kam/pkg/pipelines/resources"
+	"github.com/spf13/afero"
+	"sigs.k8s.io/yaml"
+)
+
+func TestWriteResources(t *testing.T) {
+	fs := afero.NewOsFs()
+	homeEnv := "HOME"
+	originalHome := os.Getenv(homeEnv)
+	defer os.Setenv(homeEnv, originalHome)
+	path, cleanup := makeTempDir(t)
+	defer cleanup()
+	os.Setenv("HOME", path)
+	sampleYAML := namespaces.Create("test", "https://github.com/org/test")
+	r := res.Resources{
+		"test/myfile.yaml": sampleYAML,
+	}
+
+	_, err := WriteResources(fs, "~/manifest", r)
+	assertNoError(t, err)
+
+	want, err := yaml.Marshal(sampleYAML)
+	assertNoError(t, err)
+	got, err := ioutil.ReadFile(filepath.Join(path, "manifest/test/myfile.yaml"))
+	assertNoError(t, err)
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Fatalf("files not written to correct location: %s", diff)
+	}
+}
+
+func makeTempDir(t *testing.T) (string, func()) {
+	t.Helper()
+	dir, err := ioutil.TempDir(os.TempDir(), "manifest")
+	assertNoError(t, err)
+	return dir, func() {
+		err := os.RemoveAll(dir)
+		assertNoError(t, err)
+	}
+}
+
+func assertNoError(t *testing.T, err error) {
+	t.Helper()
+	if err != nil {
+		t.Fatal(err)
+	}
+}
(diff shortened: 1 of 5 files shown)


pkg/pipelines/yaml/resources_test.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+func TestWriteResources(t *testing.T) {
+	fs := afero.NewOsFs()
+	homeEnv := "HOME"
+	originalHome := os.Getenv(homeEnv)
+	defer os.Setenv(homeEnv, originalHome)
+	path, cleanup := makeTempDir(t)
+	defer cleanup()
+	os.Setenv("HOME", path)
```
bigkevmcd, line 23: This could use homeEnv ?

bigkevmcd: we could also use a test for what happens if you don't provide `~` at the start of the filename?

bigkevmcd (changes requested):

pkg/pipelines/yaml/resources_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	want, err := yaml.Marshal(sampleYAML)
+	assertNoError(t, err)
+	got, err := ioutil.ReadFile(filepath.Join(path, "manifest/test/myfile.yaml"))
+	assertNoError(t, err)
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Fatalf("files not written to correct location: %s", diff)
+	}
+}
+
```
wtam2018, line 40: Should we call our ol friend afero here?
fakeFs := ioutils.NewMemoryFilesystem()
path := afero.GetTempDir(fakeFs, "test")

pkg/pipelines/yaml/resources_test.go
```diff
@@ -0,0 +1,56 @@
...
+	t.Helper()
+	dir, err := ioutil.TempDir(os.TempDir(), "manifest")
+	assertNoError(t, err)
+	return dir, func() {
+		err := os.RemoveAll(dir)
+		assertNoError(t, err)
+	}
+}
+
+func assertNoError(t *testing.T, err error) {
```
wtam2018, line 51: It would be nice if we can reuse this assertion.  It is duplicated from at least 3 other packages. /pkg/pipelies/dryrun, /pkg/pipelines/scm and pkg/pipelies.

pkg/pipelines/yaml/resources_test.go
bigkevmcd: Can you explain a bit more about this?

Are you proposing to mock out `path/filepath` ?

chetan-rns: @wtam2018 `homedir.Expand` in WriteResources() doesn't work with afero's in-memory filesystem. Hence the underlying fs is being used

pkg/pipelines/yaml/resources_test.go
chetan-rns: Yeah. @bigkevmcd @wtam2018 Should we consider using an external package to take care of assertions in unit tests? To maintain uniformity across the codebase

bigkevmcd: we can move that single function to `/test` in the repo?

bigkevmcd: Or, as odo does (which is slightly unusual) to `/tests/helper`

pkg/pipelines/yaml/resources_test.go
chetan-rns: Updated. I've also added a case for writing to a path without permission

pkg/pipelines/yaml/resources_test.go
chetan-rns: There's already a helper package. I've moved this function there

wtam2018:
/lgtm
@bigkevmcd  I think you have to approve it to clear the "change requested" that you put on this PR.

bigkevmcd (approved):

bigkevmcd:
@wtam2018 feels like there's a configuration failure, elsewhere I can approve through GitHub and it accepts it as an approval in Prow, but having to approve/request changes through GitHub's approval, _and_ through Prow feels like we're doing something wrong.

wtam2018:
> @wtam2018 feels like there's a configuration failure, elsewhere I can approve through GitHub and it accepts it as an approval in Prow, but having to approve/request changes through GitHub's approval, _and_ through Prow feels like we're doing something wrong.

@bigkevmcd lgtm_acts_as_approve has been enabled.    https://github.com/redhat-developer/kam/pull/8#event-3814600251 was  merged automatically with only lgtm.  This current PR had a "Change Requested" put on.  It could make affect the ci bot.   In the future, we may want to use `/hold` instead.

PR #26, opened 2020-09-24
Sealed -secret service and namespace check to trigger interactive prompt

gaganhegde:
**What type of PR is this?**
> /kind bug

The sealed secret check to prompt the ui was corrected, previously the test went undetected if not installed in the default namespace.

https://issues.redhat.com/browse/GITOPS-348

### DIFF
--- a/pkg/cmd/bootstrap.go
+++ b/pkg/cmd/bootstrap.go
@@ -16,6 +16,7 @@ import (
 	"github.com/redhat-developer/kam/pkg/pipelines/ioutils"
 	"github.com/redhat-developer/kam/pkg/pipelines/statustracker"
 	"github.com/spf13/cobra"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
@@ -160,7 +161,8 @@ func missingFlagErr(flags []string) error {
 func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client) error {
 	log.Progressf("\nStarting interactive prompt\n")
 	// ask for sealed secrets only when default is absent
-	if client.CheckIfSealedSecretsExists(defaultSealedSecretsServiceName) != nil {
+	sealedSecretServiceStatus, _ := client.CheckIfSealedSecretsExists(defaultSealedSecretsServiceName)
+	if !sealedSecretServiceStatus {
 		io.SealedSecretsService.Namespace = ui.EnterSealedSecretNamespace(&io.SealedSecretsService)
 	}
 	io.GitOpsRepoURL = utility.AddGitSuffixIfNecessary(ui.EnterGitRepo())
@@ -205,7 +207,7 @@ func checkBootstrapDependencies(io *BootstrapParameters, client *utility.Client,
 	log.Progressf("\nChecking dependencies\n")
 
 	spinner.Start("Checking if Sealed Secrets is installed with the default configuration", false)
-	err := client.CheckIfSealedSecretsExists(defaultSealedSecretsServiceName)
+	_, err := client.KubeClient.CoreV1().Services(defaultSealedSecretsServiceName.Namespace).Get(defaultSealedSecretsServiceName.Name, v1.GetOptions{})
 	setSpinnerStatus(spinner, "Please install Sealed Secrets operator from OperatorHub", err)
 	if err == nil {
 		io.SealedSecretsService.Name = sealedSecretsController
--- a/pkg/cmd/bootstrap.go
+++ b/pkg/cmd/bootstrap.go
@@ -101,25 +101,15 @@ func (io *BootstrapParameters) Complete(name string, cmd *cobra.Command, args []
 		identifier := factory.NewDriverIdentifier(factory.Mapping(host, io.PrivateRepoDriver))
 		factory.DefaultIdentifier = identifier
 	}
-	err = checkBootstrapDependencies(io, client, log.NewStatus(os.Stdout))
-	if err != nil {
+	if err := checkBootstrapDependencies(io, client, log.NewStatus(os.Stdout)); err != nil {
 		return err
 	}
-	// ask for sealed secrets only when default is absent
-	flagset := cmd.Flags()
-	if flagset.NFlag() == 0 {
-		err = initiateInteractiveMode(io, client)
-		if err != nil {
-			return err
-		}
-	} else {
-		addGitURLSuffixIfNecessary(io)
-		err := nonInteractiveMode(io, client)
-		if err != nil {
-			return err
-		}
+
+	if cmd.Flags().NFlag() == 0 {
+		return initiateInteractiveMode(io, client)
 	}
-	return nil
+	addGitURLSuffixIfNecessary(io)
+	return nonInteractiveMode(io, client)
 }
 
 func addGitURLSuffixIfNecessary(io *BootstrapParameters) {
@@ -206,8 +196,10 @@ func checkBootstrapDependencies(io *BootstrapParameters, client *utility.Client,
 	if err := client.CheckIfSealedSecretsExists(defaultSealedSecretsServiceName); err != nil {
 		warnIfNotFound(spinner, "Please install Sealed Secrets Operator from OperatorHub", err)
 		if !apierrors.IsNotFound(err) {
-			return fmt.Errorf("failed to check for sealed secrets Operator: %w", err)
+			return fmt.Errorf("failed to check for Sealed Secrets Operator: %w", err)
 		}
+		// We do not add Sealed Secret Operator to missingDeps since we this dependency can be resolved
+		// by optional flags or interactive user inputs.
 	} else {
 		io.SealedSecretsService = defaultSealedSecretsServiceName
 	}
@@ -331,14 +323,4 @@ func hostFromURL(s string) (string, error) {
 		return "", err
 	}
 	return strings.ToLower(p.Host), nil
-}
-
-//Checks if the given element is present in the array
-func contains(arr []string, str string) bool {
-	for _, a := range arr {
-		if a == str {
-			return true
-		}
-	}
-	return false
 }
--- a/pkg/cmd/bootstrap.go
+++ b/pkg/cmd/bootstrap.go
@@ -201,54 +201,49 @@ func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client) er
 }
 
 func checkBootstrapDependencies(io *BootstrapParameters, client *utility.Client, spinner status) error {
-	var errs []error
+	missingDeps := []string{}
 	log.Progressf("\nChecking dependencies\n")
 
 	spinner.Start("Checking if Sealed Secrets is installed with the default configuration", false)
-	err := client.CheckIfSealedSecretsExists(defaultSealedSecretsServiceName)
-	setSpinnerStatus(spinner, "Please install Sealed Secrets operator from OperatorHub", err)
-	if err == nil {
-		io.SealedSecretsService.Name = sealedSecretsController
-		io.SealedSecretsService.Namespace = sealedSecretsNS
-	} else if !apierrors.IsNotFound(err) {
-		return clusterErr(err.Error())
+	if err := client.CheckIfSealedSecretsExists(defaultSealedSecretsServiceName); err != nil {
+		warnIfNotFound(spinner, "Please install Sealed Secrets operator from OperatorHub", err)
+		if !apierrors.IsNotFound(err) {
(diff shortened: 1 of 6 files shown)


pkg/cmd/bootstrap.go
```diff
@@ -40,6 +40,8 @@ var supportedDrivers = drivers{
 	"gitlab",
 }
 
+var sealedSecretDefaultValue = types.NamespacedName{Namespace: sealedSecretsNS, Name: sealedSecretsController}
```
wtam2018, line 43: This is the default sealed secrets service name, right?  Rename it to something like defaultSealedSecretsServiceName?   Also, should we put it togetter with the var above?  
```
var (
    supportedDrivers = drivers{	"github", "gitlab" }
    ...
)

File: pkg/cmd/bootstrap.go [hunk 2]
```diff
@@ -146,10 +148,10 @@ func missingFlagErr(flags []string) error {
...
 // initiateInteractiveMode starts the interactive mode impplementation if no flags are passed.
-func initiateInteractiveMode(io *BootstrapParameters) error {
+func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client) error {
 	log.Progressf("\nStarting interactive prompt\n")
 	// ask for sealed secrets only when default is absent
-	if io.SealedSecretsService == (types.NamespacedName{}) {
+	if client.CheckIfSealedSecretsExists(sealedSecretDefaultValue) != nil {
 		io.SealedSecretsService.Name = ui.EnterSealedSecretService(&io.SealedSecretsService)
 	}
 	io.GitOpsRepoURL = utility.AddGitSuffixIfNecessary(ui.EnterGitRepo())
```
wtam2018, line 157: client.CheckIfSealedSecretsExists(sealedSecretDefaultValue) can return some error other than "Not found".

Also, what if the user's entered SS service does not exist?

pkg/cmd/bootstrap.go
gaganhegde: Sure William will do!

pkg/cmd/bootstrap.go
gaganhegde: Yes William thanks for bringing that up, the prompt now throws errors based on whether the namespace was missing or the service is missing. Thank you.

pkg/cmd/ui/validate.go
```diff
@@ -134,13 +134,18 @@ func validateAccessToken(input interface{}, serviceRepo string) error {
...
+		sealedSecretService.Name = EnterSealedSecretService(sealedSecretService)
+		err = client.CheckIfSealedSecretsExists(*sealedSecretService)
+		if errorType.IsNotFound(err) {
+			return fmt.Errorf("The service %s is not present in the namespace %s", sealedSecretService.Name, sealedSecretService.Namespace)
+		}
 		if err != nil {
-			if compareError(err, sealedSecretService.Name) {
-				return fmt.Errorf("The given service %q is not installed in the right namespace %q", sealedSecretService.Name, sealedSecretService.Namespace)
-			}
 			return errors.New("sealed secrets could not be configured sucessfully")
```
bigkevmcd, line 149: Perhaps we should include the actual error here, to make debugging easier?

Also, the error is misleading, we're not trying to configure sealed secrets?

pkg/cmd/bootstrap.go
bigkevmcd: @gaganhegde I'm not sure this is the right fix.

The issue before was that if it got an error checking for sealed secrets, it assumed it was because the service couldn't be found, so prompted for the name/namespace.

What happens if it can't connect to the Kubernetes server? What happens if it gets an auth failure?

Is prompting for the name/namespace the right thing to do in that case?

wtam2018 (commented):
It looks like you are still in the process of fixing the issues.  Please address the comments and notify me when it is ready for review..

pkg/cmd/ui/validate.go
```diff
@@ -136,25 +136,35 @@ func validateAccessToken(input interface{}, serviceRepo string) error {
...
+			return err
+		}
+		exists, _ := namespaces.Exists(clientSet, s)
+		if !exists {
+			return fmt.Errorf("The namespace %s is not found on the cluster", s)
+		}
+		if err != nil {
+			return err
+		}
+		sealedSecretService.Namespace = s
```
wtam2018, line 154: Do you need this here?
``
		if err != nil {
			return err
		}
        ```

pkg/cmd/ui/validate.go
```diff
@@ -136,25 +136,35 @@ func validateAccessToken(input interface{}, serviceRepo string) error {
...
+		sealedSecretService.Name = enterSealedSecretService()
+		err = client.CheckIfSealedSecretsExists(*sealedSecretService)
```
wtam2018, line 156: You need to nil check the err before use.

pkg/cmd/bootstrap.go
```diff
@@ -154,11 +159,11 @@ func missingFlagErr(flags []string) error {
 }
 
 // initiateInteractiveMode starts the interactive mode impplementation if no flags are passed.
-func initiateInteractiveMode(io *BootstrapParameters) error {
+func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client) error {
 	log.Progressf("\nStarting interactive prompt\n")
 	// ask for sealed secrets only when default is absent
-	if io.SealedSecretsService == (types.NamespacedName{}) {
-		io.SealedSecretsService.Name = ui.EnterSealedSecretService(&io.SealedSecretsService)
+	if client.CheckIfSealedSecretsExists(defaultSealedSecretsServiceName) != nil {
```
wtam2018, line 165: You can get error for a number of reasons.   Service does not exist is only one of the many possibly reasons.

pkg/cmd/bootstrap.go
```diff
@@ -154,11 +159,11 @@ func missingFlagErr(flags []string) error {
...
+		io.SealedSecretsService.Namespace = ui.EnterSealedSecretNamespace(&io.SealedSecretsService)
```
wtam2018, line 166: EnterSealedSecretNamespace is a misnomer.  namesspace and name entered by the user.

pkg/cmd/bootstrap_test.go
```diff
@@ -233,14 +227,14 @@ func TestDependenciesWithNothingInstalled(t *testing.T) {
...
-Checking if OpenShift Pipelines Operator is installed with the default configuration[Please install OpenShift Pipelines operator from OperatorHub]`
+Checking if Sealed Secrets is installed with the default configuration [Please install Sealed Secrets operator from OperatorHub]
+Checking if ArgoCD Operator is installed with the default configuration [Please install ArgoCD operator from OperatorHub]
+Checking if OpenShift Pipelines Operator is installed with the default configuration [Please install OpenShift Pipelines operator from OperatorHub]`
 
 	buff := &bytes.Buffer{}
 	fakeSpinner := &mockSpinner{writer: buff}
-	err := checkBootstrapDependencies(&BootstrapParameters{&pipelines.BootstrapOptions{}}, fakeClient, fakeSpinner)
-	wantErr := "Failed to satisfy the required dependencies"
+	_, err := checkBootstrapDependencies(&BootstrapParameters{&pipelines.BootstrapOptions{}}, fakeClient, fakeSpinner)
```
wtam2018, line 236: None of the unit tests validate the returned dependencies values.  Please add checks.

pkg/cmd/bootstrap.go
```diff
@@ -197,55 +202,50 @@ func initiateInteractiveMode(io *BootstrapParameters) error {
...
-		io.SealedSecretsService.Name = sealedSecretsController
-		io.SealedSecretsService.Namespace = sealedSecretsNS
-	} else if !apierrors.IsNotFound(err) {
-		return clusterErr(err.Error())
+	if err := client.CheckIfSealedSecretsExists(defaultSealedSecretsServiceName); err != nil {
+		warnIfNotFound(spinner, "Please install Sealed Secrets operator from OperatorHub", err)
+		if !apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to check for sealed secrets operator: %w", err)
+		}
+		missingDeps = append(missingDeps, "Sealed Secrets Operator")
```
wtam2018, line 215: Please make "Sealed Secrets Operator" a const

pkg/cmd/bootstrap.go
```diff
@@ -197,55 +202,50 @@ func initiateInteractiveMode(io *BootstrapParameters) error {
...
-	setSpinnerStatus(spinner, "Please install ArgoCD operator from OperatorHub, with an ArgoCD resource called 'argocd'", err)
-	if err != nil {
+	if err := client.CheckIfArgoCDExists(argoCDNS); err != nil {
+		warnIfNotFound(spinner, "Please install ArgoCD operator from OperatorHub", err)
 		if !apierrors.IsNotFound(err) {
-			return clusterErr(err.Error())
+			return nil, fmt.Errorf("failed to check for ArgoCD operator: %w", err)
 		}
-		errs = append(errs, err)
+		missingDeps = append(missingDeps, "ArgoCD operator")
```
wtam2018, line 226: Operator (cap)

pkg/cmd/bootstrap.go
```diff
@@ -197,55 +202,50 @@ func initiateInteractiveMode(io *BootstrapParameters) error {
...
 
 	spinner.Start("Checking if ArgoCD Operator is installed with the default configuration", false)
-	err = client.CheckIfArgoCDExists(argoCDNS)
-	setSpinnerStatus(spinner, "Please install ArgoCD operator from OperatorHub, with an ArgoCD resource called 'argocd'", err)
-	if err != nil {
+	if err := client.CheckIfArgoCDExists(argoCDNS); err != nil {
+		warnIfNotFound(spinner, "Please install ArgoCD operator from OperatorHub", err)
 		if !apierrors.IsNotFound(err) {
-			return clusterErr(err.Error())
+			return nil, fmt.Errorf("failed to check for ArgoCD operator: %w", err)
```
wtam2018, line 224: Operator (cap)

pkg/cmd/bootstrap.go
```diff
@@ -197,55 +202,50 @@ func initiateInteractiveMode(io *BootstrapParameters) error {
...
+	} else {
+		io.SealedSecretsService = defaultSealedSecretsServiceName
 	}
 
 	spinner.Start("Checking if ArgoCD Operator is installed with the default configuration", false)
-	err = client.CheckIfArgoCDExists(argoCDNS)
-	setSpinnerStatus(spinner, "Please install ArgoCD operator from OperatorHub, with an ArgoCD resource called 'argocd'", err)
-	if err != nil {
+	if err := client.CheckIfArgoCDExists(argoCDNS); err != nil {
+		warnIfNotFound(spinner, "Please install ArgoCD operator from OperatorHub", err)
```
wtam2018, line 222: Operator (cap)

pkg/cmd/bootstrap_test.go
```diff
@@ -364,7 +358,7 @@ func (m *mockSpinner) End(status bool) {
 }
 
 func (m *mockSpinner) WarningStatus(status string) {
-	fmt.Fprintf(m.writer, "[%s]", status)
```
wtam2018, line 367: Unintentinal? Revert this?

pkg/cmd/bootstrap_test.go
```diff
@@ -290,13 +284,13 @@ func TestDependenciesWithNoPipelines(t *testing.T) {
...
-Checking if OpenShift Pipelines Operator is installed with the default configuration[Please install OpenShift Pipelines operator from OperatorHub]`
+Checking if OpenShift Pipelines Operator is installed with the default configuration [Please install OpenShift Pipelines operator from OperatorHub]`
 
 	buff := &bytes.Buffer{}
 	fakeSpinner := &mockSpinner{writer: buff}
 	wizardParams := &BootstrapParameters{&pipelines.BootstrapOptions{}}
-	err := checkBootstrapDependencies(wizardParams, fakeClient, fakeSpinner)
-	wantErr := "Failed to satisfy the required dependencies"
+	_, err := checkBootstrapDependencies(wizardParams, fakeClient, fakeSpinner)
+	wantErr := "failed to satisfy the required dependencies: OpenShift Pipelines Operator"
```
wtam2018, line 293: Not sure why it starts with lowercase.

pkg/cmd/bootstrap.go
gaganhegde: Any other error will be taken care of before the interactive mode in the validateSealedSecrets()

pkg/cmd/bootstrap_test.go
bigkevmcd: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

PR #43, opened 2020-09-30
Update default values for sealed secrets in service command

chetan-rns:
**What type of PR is this?**

> /kind bug

Fixes #?

### DIFF
--- a/pkg/cmd/service/add.go
+++ b/pkg/cmd/service/add.go
@@ -74,8 +74,8 @@ func newCmdAdd(name, fullName string) *cobra.Command {
 	cmd.Flags().StringVar(&o.InternalRegistryHostname, "image-repo-internal-registry-hostname", "image-registry.openshift-image-registry.svc:5000", "Host-name for internal image registry e.g. docker-registry.default.svc.cluster.local:5000, used if you are pushing your images to the internal image registry")
 	cmd.Flags().StringVar(&o.PipelinesFolderPath, "pipelines-folder", ".", "Folder path to retrieve manifest, eg. /test where manifest exists at /test/pipelines.yaml")
 
-	cmd.Flags().StringVar(&o.SealedSecretsService.Namespace, "sealed-secrets-ns", "kube-system", "Namespace in which the Sealed Secrets operator is installed, automatically generated secrets are encrypted with this operator")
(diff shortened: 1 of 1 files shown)


pkg/cmd/service/add.go
```diff
@@ -74,8 +74,8 @@ func newCmdAdd(name, fullName string) *cobra.Command {
 	cmd.Flags().StringVar(&o.InternalRegistryHostname, "image-repo-internal-registry-hostname", "image-registry.openshift-image-registry.svc:5000", "Host-name for internal image registry e.g. docker-registry.default.svc.cluster.local:5000, used if you are pushing your images to the internal image registry")
 	cmd.Flags().StringVar(&o.PipelinesFolderPath, "pipelines-folder", ".", "Folder path to retrieve manifest, eg. /test where manifest exists at /test/pipelines.yaml")
 
-	cmd.Flags().StringVar(&o.SealedSecretsService.Namespace, "sealed-secrets-ns", "kube-system", "Namespace in which the Sealed Secrets operator is installed, automatically generated secrets are encrypted with this operator")
-	cmd.Flags().StringVar(&o.SealedSecretsService.Name, "sealed-secrets-svc", "sealed-secrets-controller", "Name of the Sealed Secrets services that encrypts secrets")
+	cmd.Flags().StringVar(&o.SealedSecretsService.Namespace, "sealed-secrets-ns", "cicd", "Namespace in which the Sealed Secrets operator is installed, automatically generated secrets are encrypted with this operator")
```
bigkevmcd, line 77: This was caused by us repeating the same logic over, we should simplify this to a simple function that adds any duplicated vars to a `*cobra.Command`

bigkevmcd (approved):

bigkevmcd:
Let's fix this in a better way in a separate branch.

pkg/cmd/service/add.go
wtam2018: @chetan-rns are we removing --sealed-secrets-svc flag?

wtam2018: There are constants `sealedSecretsController` and `sealedSecretsNS` in pkg/cmd/bootstrap and we are not using.

PR #56, opened 2020-10-06
kam service add creates a new config directory.

gaganhegde:
**What type of PR is this?**
/kind bug

The `kam service add` command does not create the `config` folder in the base directory. This fix corrects that error.

* [x] Documentation has been updated.

https://issues.redhat.com/browse/GITOPS-360

### DIFF
--- a/pkg/pipelines/environment_test.go
+++ b/pkg/pipelines/environment_test.go
@@ -35,7 +35,10 @@ func TestAddEnv(t *testing.T) {
 	}
 	for _, path := range wantedPaths {
 		t.Run(fmt.Sprintf("checking path %s already exists", path), func(rt *testing.T) {
-			assertFileExists(rt, fakeFs, filepath.Join(gitopsPath, path))
+			exists, _ := fakeFs.Exists(filepath.Join(gitopsPath, path))
+			if !exists {
+				t.Fatalf("The file is not present at path : %v", path)
+			}
 		})
 	}
 
@@ -74,7 +77,11 @@ func TestAddEnvWithClusterProvided(t *testing.T) {
 	}
 	for _, path := range wantedPaths {
 		t.Run(fmt.Sprintf("checking path %s already exists", path), func(rt *testing.T) {
-			assertFileExists(rt, fakeFs, filepath.Join(gitopsPath, path))
+			exists, _ := fakeFs.Exists(filepath.Join(gitopsPath, path))
+			if !exists {
+				t.Fatalf("The file is not present at path : %v", path)
+			}
+
 		})
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/pipelines/environment_test.go
```diff
@@ -237,6 +237,20 @@ func assertFileExists(t *testing.T, testFs afero.Fs, path string) {
 	}
 }
 
+func assertDirExists(t *testing.T, testFs afero.Fs, path string) {
+	t.Helper()
+	exists, err := afero.Exists(testFs, path)
```
bigkevmcd, line 242: You don't need to test for `Exists` and `DirExists` as they both fail if the path doesn't exist.

pkg/pipelines/service.go
```diff
@@ -195,3 +200,19 @@ func createSvcImageBinding(cfg *config.PipelinesConfig, env *config.Environment,
 	resourceFilePath := makeImageBindingPath(cfg, filename)
 	return name, filename, res.Resources{resourceFilePath: triggers.CreateImageRepoBinding(cfg.Name, name, imageRepo, strconv.FormatBool(isTLSVerify))}
 }
+
+func createConfigFolder(m *config.Manifest, appFs afero.Fs, o *AddServiceOptions) error {
+	basePath, err := homedir.Expand(o.PipelinesFolderPath)
+	if err != nil {
+		return err
```
bigkevmcd, line 207: how about expanding on this error a bit?

Getting the error [`cannot expand user-specific home dir`](https://github.com/mitchellh/go-homedir/blob/master/homedir.go#L68) probably isn't that useful to a user?

bigkevmcd (changes requested):
I wonder if we shouldn't simplify `pkg/pipelines/ioutils/file_utils.go` to something like:

```go
func NewFilesystem() afero.Afero {
       return afero.Afero{Fs: afero.NewOsFs()}
}
```
Which would arguably improve this code to something like:

```go
isDir, err := testFs.DirExists(path)
```
What do you think?

wtam2018:
Please update the doc.  Remove the node about create `config` folder if it does not exist.

bigkevmcd (approved):

wtam2018:
/hold
@gaganhegde  please update docs.

pkg/pipelines/service.go
```diff
@@ -195,3 +200,19 @@ func createSvcImageBinding(cfg *config.PipelinesConfig, env *config.Environment,
...
+	basePath, err := homedir.Expand(o.PipelinesFolderPath)
+	if err != nil {
+		return fmt.Errorf("Cannot expand the pipelines.yaml path : %s", o.PipelinesFolderPath)
+	}
+	env := m.GetEnvironment(o.EnvName)
+	app := m.GetApplication(o.EnvName, o.AppName)
+	servicePath := config.PathForService(app, env, o.ServiceName)
+	finalPath := filepath.Join(basePath, servicePath, "base", "config")
+	err = appFs.MkdirAll(finalPath, 0755)
+	if err != nil {
```
wtam2018, line 214: Will this fail if directory already exists?

gaganhegde: No William, it wont. if the directory is not present it will create one.

bigkevmcd: https://golang.org/pkg/os/#MkdirAll

PR #63, opened 2020-10-14
add kam binary to gitignore

wtam2018:
**What type of PR is this?**

> /kind bug
/kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

Add `kam` binary to .gitignore to prevent it from accidentally getting included in commits.

Fixes #?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 # vendor/
 
 # Ignore binary
+kam
 ./kam
 kam_windows_amd64.exe
 kam_linux_amd64
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@
 
 # Ignore binary
 kam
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -15,7 +15,7 @@
 # vendor/
 
 # Ignore binary
-./kam
+kam
```
bigkevmcd, line 18: This doesn't do what you think it does.

It's matching on anything with `kam` in the name, which means that it will ignore files with kam in them.

bigkevmcd (changes requested):

.gitignore
bigkevmcd: (I fixed this when the newly generated documentation files weren't picked up).

PR #67, opened 2020-10-20
Fixing doc generate failure

amitkrout:
**What type of PR is this?**

/kind failing-test
/kind enhancement

Prow unable to sense github action failure. So to avoid merges after failure, i have added the `make docs` in the prow job so that the failure status will be identified and pr won't merge into the master. 

Screen Shot 2020-10-20 at 8 13 33 PM

Fixes NA

`make docs` error should be reported in prow job

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ openshiftci-presubmit-unittests:
 
 .PHONY: cmd-docs
 cmd-docs:
-	go run tools/cmd-docs/main.go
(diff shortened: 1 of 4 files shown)


amitkrout:
> In test-platform channel I will report this behavior of prow why it does not honor other CI failure even after enabling the branch level protection.

There is an open pr for the fix - https://github.com/kubernetes/test-infra/pull/19520

wtam2018 (commented):
/lgtm

PR #68, opened 2020-10-21
Create developer login session in CI

amitkrout:
**What type of PR is this?**
/kind enhancement

Providing a developer login session to run the following operations 

kam CLI bootstraps and generates CI/CD pipelines with starter app, push to gitops repository, oc apply -f /config/argocd (i.e. create argocd apps in the generated config) etc...

Fixes NA

```
+ oc login -u developer -p developer
Login successful.

You have access to the following projects and can switch between them with 'oc project <projectname>':

  * argocd
    cicd

Using project "argocd".
+ oc whoami
developer
```

Check the prow log - https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-developer_kam/68/pull-ci-redhat-developer-kam-master-v4.5-integration-e2e/1318884090304270336/artifacts/integration-e2e/integration-e2e-steps/container-logs/test.log

### DIFF
--- a/scripts/openshiftci-presubmit-all-tests.sh
+++ b/scripts/openshiftci-presubmit-all-tests.sh
@@ -16,4 +16,24 @@ export PATH="$PATH:$(pwd)/bin"
 export ARTIFACTS_DIR="/tmp/artifacts"
 export CUSTOM_HOMEDIR=$ARTIFACTS_DIR
 
+# Copy kubeconfig to temporary kubeconfig file and grant
+# read and Write permission to temporary kubeconfig file
+TMP_DIR=$(mktemp -d)
+cp $KUBECONFIG $TMP_DIR/kubeconfig
+chmod 640 $TMP_DIR/kubeconfig
+export KUBECONFIG=$TMP_DIR/kubeconfig
+
+# Login as developer
+oc login -u developer -p developer
+
+# Check login user name for debugging purpose
+oc whoami
+login_user=`oc whoami`
+if [[ $login_user == *"developer"* ]]; then
+    echo "Login to the cluster as a developer user"
+else
+    echo "Fail to login as a developer user"
+    exit 1
+fi
+
 kam version


scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -12,4 +12,24 @@ export PATH="$PATH:$(pwd)"
 export ARTIFACTS_DIR="/tmp/artifacts"
 export CUSTOM_HOMEDIR=$ARTIFACTS_DIR
```
wtam2018, line 14: `openshift-presubmit-all-tests.sh` should follow the same invocation pattern as `make openshiftci-presubmit-unittests`   That is, 
* In ci-operator, call `make openshiftci-presubmit-unittests`
* In Makefile, call `openshift-presubmit-all-tests.sh` 
* `openshift-presubmit-all-tests.sh`  source `prepare-test-clusher.sh` and there probably should not be a `prepare-test-cluster` target in the Makefile.

amitkrout: @wtam2018 I did not realize it, you have a good point that the change should be in sync and make sense too. However i can refine it little bit, for example 

```openshift-presubmit-all-tests.sh``` and ```openshiftci-presubmit-unittests.sh``` both are written for CI operation, so we should not expose these two file in the make target as it gives a wrong message that these target can run locally. But prepare-test-cluster.sh design for both local and CI run, hence its better to have a make target for user who want to configure the cluster locally through ```make prepare-test-cluster``` before running e2e tests.

Task need to be done are

- [ ] Call the script ```openshiftci-presubmit-unittests``` directly in ci-operator 
- [ ] Remove the target ```make openshiftci-presubmit-unittests``` from make file

I will fix these task in separate prs

**Review by wtam2018 — COMMENTED** [review]
fyi, you need to adjust PATH variable when this  https://github.com/redhat-developer/kam/pull/69 is landed.

File: scripts/openshiftci-presubmit-all-tests.sh (hunk 1 above)
Comment (wtam2018) [reply scripts/openshiftci-presubmit-all-tests.sh:14]: My problem is `make prepare-test-cluster` setting some environment variables like KUBECONFIG.   I don't expect make target to set them.  I'd prefer sourcing prepare-test-cluster.sh in this script or refactoring the environments to a source I can source.

Comment (wtam2018) [reply scripts/openshiftci-presubmit-all-tests.sh:14]: Having ci-operator calling directly to scripts/openshiftci-presubmit-unittests.sh has downsides.  A person looks at the scripts folder will have no ideas if this script is being used.   I don't know about making a make target is giving a wrong message.   Makefile targets serve as interface to ci-operator.

Comment (amitkrout) [reply scripts/openshiftci-presubmit-all-tests.sh:14]: > My problem is make prepare-test-cluster setting some environment variables like KUBECONFIG. I don't expect make target to set them. I'd prefer sourcing prepare-test-cluster.sh in this script or refactoring the environments to a source I can source.

We need to do it through the make target because this target supports both CI and local configuration. I have already documented the steps in https://github.com/redhat-developer/kam/blob/master/docs/test/test-guide.md#prerequisites-for-openshift-cluster for local installation. make target looks pretty straight forward and simple, in case of sourcing it in a separate file will only add an extra wrapper around the same operation. Ofcourse you can overwrite the environment variable by exporting it in the current session 

In case of cluster configuration in CI the same target will be helpful as in the script CI checks and default environment variables are properly maintained.

Comment (amitkrout) [reply scripts/openshiftci-presubmit-all-tests.sh:14]: > Having ci-operator calling directly to scripts/openshiftci-presubmit-unittests.sh has downsides. A person looks at the scripts folder will have no ideas if this script is being used. I don't know about making a make target is giving a wrong message. Makefile targets serve as interface to ci-operator.

It will be documented here at - https://github.com/redhat-developer/kam/blob/master/docs/test/ci-guide.md#ci-reference, i mean which files are responsible for CI configuration. And also a comment in the script file for example ```This is a openshift CI configuration file``` would really be helpful instead of calling it through a make target.

Comment (wtam2018) [reply scripts/openshiftci-presubmit-all-tests.sh:14]: So, if I already prepared a test cluster and close my termainal, logout or whatever I do to lose my environment.   I now want to login to the test cluster to troubleshoot.   How to I set my environment back?   I can't run "make prepared- clustter" because I don't need to re-create all these settings.  I just need my environmet and login as developer or even kubeadmin.

Comment (amitkrout) [reply scripts/openshiftci-presubmit-all-tests.sh:18]: @wtam2018 make it or source it, either way we are going to loose the environment if session is closed or logout. In case the user closing the terminal then he/she has to login the cluster session again using ```oc login -u developer -p developer```, no other configuration or environment is needed because the env and other variable used in the script is for setting configuration for first and last time. BTW if you run the traget "make prepare-test-cluster" again it repeats the same configuration and if it finds the same existing configuration does not fail  and give you back the same lost session again.

Comment (wtam2018) [reply scripts/openshiftci-presubmit-all-tests.sh:18]: @amisevsk My point is not that sourcing a script (as opposed to make target) will make your environment variables survive after logging out.   Of cause, not.   My point is if you have to re-login, you have source these environment in `prepare-test-cluster.sh` which is kicked by your `make prepare-test-cluster.sh`.
```HTPASSWD_FILE="./htpass"
USERPASS="developer"
HTPASSWD_SECRET="htpasswd-secret"
SETUP_OPERATORS="./scripts/setup-operators.sh"
# Overrideable information
DEFAULT_INSTALLER_ASSETS_DIR=${DEFAULT_INSTALLER_ASSETS_DIR:-$(pwd)}
KUBEADMIN_USER=${KUBEADMIN_USER:-"kubeadmin"}
KUBEADMIN_PASSWORD_FILE=${KUBEADMIN_PASSWORD_FILE:-"${DEFAULT_INSTALLER_ASSETS_DIR}/auth/kubeadmin-password"}
```
Whe you said, "no other environment is needed" is not true.
Now, I have to run the make target to login.    I could just source the environment and login.  Why do I have to run make xxxx and execute something the following again?
```
# Create the namespace for operator installation namespace
for i in `echo $OPERATOR_NAMESPACES`; do
    # create the namespace
    oc new-project $i
    # Let developer user have access to the project
    oc adm policy add-role-to-user edit developer
done
```
And that was my point.

amitkrout: > These are just my concerns. If you are confident with your approach, I am fine with that. To login, I only need KUBEADMIN_USER, KUBEADMIN_PASSWORD_FILE, KUBECONFIG. They are small enough for me to set manually.

@wtam2018 Thanks for the details. I have documented how to prepare cluster locally to run e2e test in the ```NOTE``` section of - https://github.com/redhat-developer/kam/blob/master/docs/test/test-guide.md#prerequisites-for-openshift-cluster. And yes for the lost session i will add few more steps for how to get it by setting up env variable ```KUBEADMIN_USER, KUBEADMIN_PASSWORD_FILE, KUBECONFIG``` manually without kicking up ```make prepare-test-cluster.sh``` again.

PR #69, opened 2020-10-21
output built binaries in  "dist" folder

wtam2018:
**What type of PR is this?**
/kind enhancement

This PR changes the location of where binaries are outputted.   Instead of putting the built binaries (kam, etc) in the top level folder, they are not put in the "dist" folder.   .gitIngore file is updated to ignore files in "dist" folder.

Fixes #?

* Remove any `kam` binary at the top level folder.
* Run `make bin` to compile `kam`
* Check built file `ls -l dist`
* Run `make clean` 
* Check binaries are removed `ls -l dist`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,7 +33,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }} # This pulls from the CREATE RELEASE step above, referencing it's ID to get its outputs object, which include a `upload_url`. See this blog post for more info: https://jasonet.co/posts/new-features-of-github-actions/#passing-data-to-future-steps 
-          asset_path: ./kam_linux_amd64
+          asset_path: ./dist/kam_linux_amd64
           asset_name: kam_linux_amd64
           asset_content_type: application/octet-stream
       - name: Upload kam_windows_amd64.exe
@@ -43,7 +43,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }} # This pulls from the CREATE RELEASE step above, referencing it's ID to get its outputs object, which include a `upload_url`. See this blog post for more info: https://jasonet.co/posts/new-features-of-github-actions/#passing-data-to-future-steps 
-          asset_path: ./kam_windows_amd64.exe
+          asset_path: ./dist//kam_windows_amd64.exe
           asset_name: kam_windows_amd64.exe
           asset_content_type: application/octet-stream          
       - name: Upload kam_darwin_amd64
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }} # This pulls from the CREATE RELEASE step above, referencing it's ID to get its outputs object, which include a `upload_url`. See this blog post for more info: https://jasonet.co/posts/new-features-of-github-actions/#passing-data-to-future-steps 
-          asset_path: ./kam_windows_amd64.exe
+          asset_path: ./dist//kam_windows_amd64.exe
```
bigkevmcd, line 46: Why two `//` here?

Makefile
```diff
@@ -55,7 +56,7 @@ test:
 
 .PHONY: clean
 clean:
-	@rm -f $(WINDOWS) $(LINUX) $(DARWIN) ${EXECUTABLE} 
+	@rm -f $(DIST_DIR)/$(WINDOWS) $(DIST_DIR)/$(LINUX) $(DIST_DIR)/$(DARWIN) $(DIST_DIR)/${EXECUTABLE}
```
bigkevmcd, line 59: this could be simplified to

```
@rm -f $(DIST_DIR)/*
```

bigkevmcd (changes requested):

bigkevmcd (dismissed):
Looks good, thanks!

.gitignore
```diff
@@ -15,7 +15,7 @@
 # vendor/
 
 # Ignore binary
-./kam
+dist/
 kam_windows_amd64.exe
 kam_linux_amd64
 kam_darwin_amd64
```
amitkrout, line 21: Adding of individual binary is not required if you are adding the parent directory, so remove the distribution binaries

amitkrout (changes requested):

Makefile
```diff
@@ -43,7 +44,7 @@ gofmt:
 
 .PHONY: bin
 bin:
-	go build  ${CFLAGS} -o $(EXECUTABLE) -ldflags=$(LD_FLAGS) cmd/kam/kam.go 
+	go build  ${CFLAGS} -o $(DIST_DIR)/$(EXECUTABLE) -ldflags=$(LD_FLAGS) cmd/kam/kam.go
```
amitkrout, line 47: AFAIK this is not a part of binary distribution. Its a regular binary creation on the platform you are executing, for example on linux platform it produce linux compiled bit and on windows it produce .exe file. I think it does not make sense to add this in distribution directory.

amitkrout (changes requested):

scripts/openshiftci-presubmit-unittests.sh
```diff
@@ -29,6 +29,6 @@ fi
 # crosscompile and publish artifacts
 make all_platforms
 
-cp kam_darwin_amd64 $CUSTOM_HOMEDIR
-cp kam_linux_amd64 $CUSTOM_HOMEDIR
-cp kam_windows_amd64.exe $CUSTOM_HOMEDIR
+cp dist/kam_darwin_amd64 $CUSTOM_HOMEDIR
+cp dist/kam_linux_amd64 $CUSTOM_HOMEDIR
+cp dist/kam_windows_amd64.exe $CUSTOM_HOMEDIR
```
amitkrout, line 34: 👍

Makefile
wtam2018: what would you suggest?

bigkevmcd: how about we split this into `make dist` and `make bin` ?

PR #71, opened 2020-10-23
Adding argocd install script that can be referenced from other E2E tests

dewan-ahmed:
**What type of PR is this?**
/kind enhancement

This PR adds a script to install ArgoCD on darwin/linux that's needed as part of GITOPS-333 issue.

Documentation will be updated once the entire work is completed for GITOPS-333.

[GITOPS-333](https://issues.redhat.com/projects/GITOPS/issues/GITOPS-333) partial.

Execute this script on mac/linux where you don't have ArgoCD installed. The script will install ArgoCD CLI. Since the ArgoCD server address is not set, you might see an error at the end. But the goal of this script is to install ArgoCD CLI.

### DIFF
--- a/scripts/install-argocd.sh
+++ b/scripts/install-argocd.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-set -e
 
 # Find the latest ArgoCD version
 VERSION=$(curl --silent "https://api.github.com/repos/argoproj/argo-cd/releases/latest" | grep '"tag_name"' | sed -E 's/.*"([^"]+)".*/\1/')
--- a/scripts/install-argocd.sh
+++ b/scripts/install-argocd.sh
@@ -9,7 +9,7 @@ PLATFORM=`uname -s | awk '{print tolower($0)}'`
 echo $PLATFORM 
 
 # Fetch the appropriate OS binary for ArgoCD
-curl -sSL -o argocd https://github.com/argoproj/argo-cd/releases/download/$VERSION/argocd-$PLATFORM-amd64
+curl -sSL -o bin/argocd https://github.com/argoproj/argo-cd/releases/download/$VERSION/argocd-$PLATFORM-amd64
 
 # Make the argocd CLI executable
 chmod +x argocd
--- a/scripts/install-argocd.sh
+++ b/scripts/install-argocd.sh
@@ -1,29 +1,16 @@
 #!/bin/bash
 set -e
 
-if [[ "$ARGOCD_SERVER" == "" ]]; then
- echo "ArgoCD server address unspecified!"
- exit 1
-fi
-
-echo "Removing previous argocd installation from path (if any)..."
-rm /usr/local/bin/argocd
-
-echo "Starting argocd installation..."
-
-# Not considering the Windows binary for now
-OS_NAME=$(uname -s)
-
 # Find the latest ArgoCD version
 VERSION=$(curl --silent "https://api.github.com/repos/argoproj/argo-cd/releases/latest" | grep '"tag_name"' | sed -E 's/.*"([^"]+)".*/\1/')
+echo $VERSION
+
+# Platform check
+PLATFORM=`uname -s | awk '{print tolower($0)}'`
+echo $PLATFORM 
 
 # Fetch the appropriate OS binary for ArgoCD
-
-curl -sSL -o /usr/local/bin/argocd https://github.com/argoproj/argo-cd/releases/download/$VERSION/argocd-$OS_NAME-amd64
+curl -sSL -o argocd https://github.com/argoproj/argo-cd/releases/download/$VERSION/argocd-$PLATFORM-amd64
 
 # Make the argocd CLI executable
-
-chmod +x /usr/local/bin/argocd
-
-# Simple check to see if Argocd binary works
-argocd version
+chmod +x argocd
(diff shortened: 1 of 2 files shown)


scripts/install-argocd.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -e
+
```
amitkrout, line 3: @dewan-ahmed I have removed few script that was doing local installation check. Let do the installation for CI first, later we can leverage it to local installation. The script looks good to me, however you missed one step of converting the Upper case to lower case. And no need to put the binary in the default linux path because this script need to be be sourced after ```make bin``` - https://github.com/redhat-developer/kam/blob/master/scripts/openshiftci-presubmit-all-tests.sh#L10. 
```
# Find the latest ArgoCD version
VERSION=$(curl --silent "https://api.github.com/repos/argoproj/argo-cd/releases/latest" | grep '"tag_name"' | sed -E 's/.*"([^"]+)".*/\1/')
echo $VERSION

# Platform check
PLATFORM=`uname -s | awk '{print tolower($0)}'`
echo $PLATFORM 

# Fetch the appropriate OS binary for ArgoCD
curl -sSL -o argocd https://github.com/argoproj/argo-cd/releases/download/$VERSION/argocd-$PLATFORM-amd64

# Make the argocd CLI executable
chmod +x argocd 
```

amitkrout (changes requested):

scripts/install-argocd.sh
dewan-ahmed: @amitkrout thanks for the feedback and fix. I implemented your feedback and tested on darwin & linux.

amitkrout:
platform issue 
```
error: some steps failed:
  * could not initialize namespace: could not create role binding for: rolebindings.rbac.authorization.k8s.io is forbidden: User "system:serviceaccount:ci:ci-operator" cannot create resource "rolebindings" in API group "rbac.authorization.k8s.io" in the namespace "ci-op-bshsp2dw" 
```
/retest

amitkrout:
Seems you accidentally commit the ```argocd``` binary. Please remove it.

amitkrout (changes requested):
After ```make bin``` in file https://github.com/redhat-developer/kam/blob/master/scripts/openshiftci-presubmit-all-tests.sh add these two lines 
```
INSTALL_ARGOCD="./scripts/install-argocd.sh"
sh $INSTALL_ARGOCD
```

dewan-ahmed:
@amitkrout ah sorry committed the binary by mistake; removed it.

Added the lines under https://github.com/redhat-developer/kam/blob/master/scripts/openshiftci-presubmit-all-tests.sh

Thanks.

scripts/install-argocd.sh
```diff
@@ -0,0 +1,15 @@
...
+# Find the latest ArgoCD version
+VERSION=$(curl --silent "https://api.github.com/repos/argoproj/argo-cd/releases/latest" | grep '"tag_name"' | sed -E 's/.*"([^"]+)".*/\1/')
+echo $VERSION
+
+# Platform check
+PLATFORM=`uname -s | awk '{print tolower($0)}'`
+echo $PLATFORM 
+
+# Fetch the appropriate OS binary for ArgoCD
+curl -sSL -o argocd https://github.com/argoproj/argo-cd/releases/download/$VERSION/argocd-$PLATFORM-amd64
```
amitkrout, line 12: Pr https://github.com/redhat-developer/kam/pull/69 created a bin dir for kam binary. Lets put the argocd binary in same directory to avoid another PATH set for argocd in ```openshiftci-presubmit-all-tests.sh```.

```curl -sSL -o bin/argocd https://github.com/argoproj/argo-cd/releases/download/$VERSION/argocd-$PLATFORM-amd64```

**Review by amitkrout — CHANGES_REQUESTED** [review]

File: scripts/install-argocd.sh [hunk 3]
```diff
@@ -0,0 +1,15 @@
...
+
+# Make the argocd CLI executable
+chmod +x argocd
```
amitkrout, line 15: ```chmod +x bin/argocd```

amitkrout (changes requested):

scripts/install-argocd.sh
dewan-ahmed: For the `install-argocd.sh` script, the binary is downloaded into current directory so shouldn't the chmod be executed on the filename for the working directory and not `bin/...`? argocd binary is not under `bin/` for the context of `install-argocd.sh` script.

amitkrout: As the pr #69 has been merged, so we need to download the argocd binary into ```bin/``` directory like i mentioned in https://github.com/redhat-developer/kam/pull/71#discussion_r511072998 with the reason. So the execute permission will be applied on ```bin/argocd```

dewan-ahmed: Thanks for the explanation @amitkrout

amitkrout:
```
 chmod: cannot access 'argocd': No such file or directory
error: failed to execute wrapped command: exit status 1 
```
Details log - https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/redhat-developer_kam/71/pull-ci-redhat-developer-kam-master-v4.5-integration-e2e/1320186431414472704#1:build-log.txt%3A905

@dewan-ahmed Please update the path here too. For example 
```chmod +x bin/argocd```

amitkrout (changes requested):

dewan-ahmed:
Platform issue:

`level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply Terraform: failed to complete the change"`

/retest

PR #82, opened 2020-11-12
Adds the prefix to the config map for the gitops-operator.

gaganhegde:
**What type of PR is this?**
kind enhancement

Adds the prefix to the config map called gitops-config in the gitops-operators namespace for the gitops operator.

There is no issue on the Jira currently.

Fixes #?


bigkevmcd (commented):
This looks ok technically, but I have a lingering concern that this is not the right thing to do.

gaganhegde:
Currently wont be in use until sealed-secrets-operator is fixed

PR #91, opened 2020-11-24
Export GitHub token

amitkrout:
**What type of PR is this?**
/kind enhancement

Github token is needed for kam day 1 operation

Fixes part of https://issues.redhat.com/browse/GITOPS-333

Token should be exported in the test container

### DIFF
--- a/scripts/openshiftci-presubmit-all-tests.sh
+++ b/scripts/openshiftci-presubmit-all-tests.sh
@@ -43,4 +43,5 @@ fi
 kam version
 
 # Providing github.com login instance
-sudo gh auth login --with-token < $KAM_GITHUB_TOKEN
+echo $HOME
+gh auth login --with-token < $KAM_GITHUB_TOKEN
--- a/scripts/openshiftci-presubmit-all-tests.sh
+++ b/scripts/openshiftci-presubmit-all-tests.sh
@@ -6,6 +6,10 @@ set -e
 set -x
 
 export CI="prow"
+export GITHUB_TOKEN=$KAM_GITHUB_TOKEN
+cd /var/run/kam-data/user-secret/
+cat secret.txt
+cd -
 make prepare-test-cluster
 make bin
 
@@ -42,6 +46,6 @@ fi
 # kam version check
 kam version
 
-# Providing github.com login instance
-echo $HOME
-gh auth login --with-token < $KAM_GITHUB_TOKEN
+# # Providing github.com login instance
+# echo $HOME
+# gh auth login --with-token < $KAM_GITHUB_TOKEN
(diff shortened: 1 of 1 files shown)


scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -39,4 +43,5 @@ else
     exit 1
 fi
 
+# kam version check
```
bigkevmcd, line 46: What's this comment indicating?

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -2,9 +2,13 @@
 
 # fail if some commands fails
 set -e
+
+# Do not show token in CI log
+set +x
+export GITHUB_TOKEN=`cat $KAM_GITHUB_TOKEN`
```
bigkevmcd, line 8: This implies that `KAM_GITHUB_TOKEN` is probably badly named, maybe something like `KAM_GITHUB_TOKEN_FILE` ?

amitkrout: Pr https://github.com/openshift/release/pull/13832 will update the new name. PTAL and apply `/lgtm` label.

scripts/openshiftci-presubmit-all-tests.sh
amitkrout: I just add a comment line for `kam version` command

bigkevmcd: But it's not checking the version?

amitkrout: Its checking the existence of the correct kam binary in the PATH by running the version command. I can re phrase to a better comment for example ```Checking kam binary in the PATH through version command.``` ?

bigkevmcd: How about something like this?

```bash
# assert that kam is on the path
```

bigkevmcd (approved):

scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -2,9 +2,13 @@
 
 # fail if some commands fails
 set -e
+
+# Do not show token in CI log
+set +x
+export GITHUB_TOKEN=`cat $KAM_GITHUB_TOKEN_FILE`
```
wtam2018, line 8: Where is `KAM_GITHUB_TOKEN_FILE`defined?

amitkrout: https://github.com/openshift/release/blob/master/ci-operator/config/redhat-developer/kam/redhat-developer-kam-master__v4.5.yaml#L42

PR #93, opened 2020-11-25
Generated argocd application to use overlays folder instead of base

iam-veeramalla:
**What type of PR is this?**
> /kind bug

Generated argocd application to use overlays folder instead of base. Given that we have an overlays directory, which is for users to override the base, we should be using the overlays directory, otherwise nothing that is put into the overlays will be picked up.

NA 

NA

Run `kam bootstrap` and see if the generated argocd application (/config/argcod/<env>-<app>-app,xml) spec/source/path uses "overlays" folder instead of "base".

### DIFF
--- a/pkg/pipelines/argocd/argocd_test.go
+++ b/pkg/pipelines/argocd/argocd_test.go
@@ -1,6 +1,7 @@
 package argocd
 
 import (
+	"path/filepath"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -35,6 +36,9 @@ var (
 			testApp,
 		},
 	}
+
+	testEnvPath     = filepath.Join(config.PathForEnvironment(testEnv), "env")
+	testEnvBasePath = filepath.Join(testEnvPath, "overlays")
 )
 
 func TestBuildCreatesArgoCD(t *testing.T) {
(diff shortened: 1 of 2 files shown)


wtam2018 (commented):
Please sync to the latest.   `makeSource` is now `makeAppSource` and you will need to fix `makeEnvSource` as well.

bigkevmcd (changes requested):
This isn't yet complete, there are other points where overlays are being used in the ArgoCD configuration.

iam-veeramalla:
> Please sync to the latest. `makeSource` is now `makeAppSource` and you will need to fix `makeEnvSource` as well.

Made changes accordingly and modified unit tests to catch any future modifications.

iam-veeramalla:
> This isn't yet complete, there are other points where overlays are being used in the ArgoCD configuration.

@bigkevmcd made the required changes

pkg/pipelines/argocd/argocd_test.go
```diff
@@ -70,7 +77,10 @@ func TestBuildCreatesArgoCD(t *testing.T) {
 			TypeMeta:   applicationTypeMeta,
 			ObjectMeta: meta.ObjectMeta(meta.NamespacedName(ArgoCDNamespace, "test-dev-http-api")),
 			Spec: argoappv1.ApplicationSpec{
-				Source: *makeAppSource(testEnv, testEnv.Apps[0], testRepoURL),
+				Source: argoappv1.ApplicationSource{
+					RepoURL: testRepoURL,
+					Path:    filepath.Join(config.PathForApplication(testEnv, testApp), "overlays"),
```
bigkevmcd, line 82: I'd probably have simplified both of these to using the same pattern, but this works for me.

Issue #96, opened 2020-11-25
Cannot run `kam bootstrap` from Users $HOME directory in Mac.


iam-veeramalla:
**Describe the bug**
`kam bootstrap` returns the below error when executed from Users $HOME directory

```
 ✗  failed to create the gitops repository: "https://github.com/iam-veeramalla/gitops.git": failed to push bootstrapped resources: failed to add files to repository in "." "warning: could not open directory 'Library/Application Support/CallHistoryTransactions/': Operation not permitted\nwarning: could not open directory 'Library/Application Support/com.apple.sharedfilelist/': Operation not permitted\nwarning: could not open directory 'Library/Application Support/com.apple.TCC/': Operation not permitted\nwarning: could not open directory 'Library/Application Support/FileProvider/': Operation not permitted\nwarning: could not open directory 'Library/Application Support/CallHistoryDB/': Operation not permitted\nwarning: could not open directory 'Library/Autosave Information/': Operation not permitted\nwarning: could not open directory 'Library/IdentityServices/': Operation not permitted\nwarning: could not open directory 'Library/Messages/': Operation not permitted\nwarning: could not open directory 'Library/Sharing/': Operation not permitted\nwarning: could not open directory 'Library/Mail/': Operation not permitted\nwarning: could not open directory 'Library/Safari/': Operation not permitted\nwarning: could not open directory 'Library/Suggestions/': Operation not permitted\nwarning: could not open directory 'Library/Containers/com.apple.archiveutility/': Operation not permitted\nwarning: could not open directory 'Library/Containers/com.apple.Safari/': Operation not permitted\nwarning: could not open directory 'Library/PersonalizationPortrait/': Operation not permitted\nwarning: could not open directory 'Library/Metadata/CoreSpotlight/': Operation not permitted\nwarning: could not open directory 'Library/VoiceTrigger/SAT/': Operation not permitted\nwarning: could not open directory 'Library/Cookies/': Operation not permitted\nwarning: could not open directory 'Library/Caches/CloudKit/': Operation not permitted\nwarning: could not open directory '.Trash/': Operation not permitted\nerror: open(\"Library/Google/GoogleSoftwareUpdate/Stats/Keystone.stats\"): Permission denied\nerror: unable to index file 'Library/Google/GoogleSoftwareUpdate/Stats/Keystone.stats'\nfatal: adding files failed\n": exit status 128
```

**To Reproduce**
[...]

PR #97, opened 2020-11-25
Restrict adding files to just the generated ones.

bigkevmcd:
**What type of PR is this?**

/kind bug

Only add generated files to the git repository when bootstrapping.

Fixes #?

#96 


bigkevmcd:
This does raise a question, should we change the default to something other than `.` perhaps `./gitops` ?

PR #109, opened 2020-12-07
Interactive command validation/ clean up

gaganhegde:
**What type of PR is this?**
> /kind cleanup

The follwing url :https://github.com/username/repo/.git leads to an error although its an acceptable url. There can be many such anomalies. This PR validates the users input with the git client and throws an error in the event of a failed authentication.

The differentiation is done based on the response from the server, if the git repo is not found the user is warned of its absence and is later asked if he wishes to push/create repo.

If the git host access token is invalid the user is prompted for the url and the host token again, incorrect tokens will error out in non-interactive mode.

The url https://github.com/username/repo/.git will then be validated to  https://github.com/username/repo.git in the command layer.

* [] Documentation has been updated.

Current implementation might not need documentation fixes since its an improvement of current implementation.

https://issues.redhat.com/browse/GITOPS-492

### DIFF
--- a/pkg/cmd/ui/ui.go
+++ b/pkg/cmd/ui/ui.go
@@ -5,28 +5,57 @@ import (
 	"net/url"
 	"path/filepath"
 
+	"github.com/openshift/odo/pkg/log"
+
 	"gopkg.in/AlecAivazis/survey.v1"
 
+	"github.com/redhat-developer/kam/pkg/cmd/utility"
 	"github.com/redhat-developer/kam/pkg/pipelines/ioutils"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 )
 
-// EnterGitRepo allows the user to specify the git repository in a prompt.
-func EnterGitRepo() string {
-	var gitOpsURL string
-	prompt := &survey.Input{
-		Message: "Provide the URL for your GitOps repository",
-		Help:    "The GitOps repository stores your GitOps configuration files, including your Openshift Pipelines resources for driving automated deployments and builds.  Please enter a valid git repository e.g. https://github.com/example/myorg.git",
-	}
-	err := survey.AskOne(prompt, &gitOpsURL, survey.Required)
-	handleError(err)
+// RepoParams struct contains the parameters required to authenticate the url with the access-token
+type RepoParams struct {
+	RepoInfo                repoInfo
+	GitHostAccessToken      string
+	KeyringSVC              bool
+	TokenRepoMatchCondition bool
+	TokenAuthenticated      bool
+}
 
-	p, err := url.Parse(gitOpsURL)
+type repoInfo struct {
+	RepoURL      string
+	RepoType     string
+	GitRepoValid bool
+}
+
+const (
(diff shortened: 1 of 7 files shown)


pkg/cmd/ui/ui.go
```diff
@@ -5,28 +5,55 @@ import (
...
+)
+
 // EnterGitRepo allows the user to specify the git repository in a prompt.
-func EnterGitRepo() string {
-	var gitOpsURL string
+func EnterGitRepo(repoType string) string {
+	var repoURL string
+	var help string
+	switch repoType {
+	case "service":
```
wtam2018, line 41:
```suggestion
	case ServiceRepoType:
```

pkg/cmd/ui/ui.go
```diff
@@ -5,28 +5,55 @@ import (
...
+		help = "The repository name where the source code of your service is situated, this will configure a very basic CI for this repository using OpenShift pipelines."
+	case "gitops":
```
wtam2018, line 43:
```suggestion
	case GitopsRepoType:
```

pkg/cmd/ui/ui.go
```diff
@@ -133,29 +160,47 @@ func enterSealedSecretService() string {
...
 // token to access upstream git hosts.
-func EnterGitHostAccessToken(serviceRepo string) string {
+func EnterGitHostAccessToken(serviceRepo string) (string, error) {
 	var accessToken string
 	prompt := &survey.Password{
 		Message: fmt.Sprintf("Please provide a token used to authenticate requests to %q", serviceRepo),
 		Help:    "Tokens are required to authenticate to git provider various operations on git repository (e.g. enable automated creation/push to git-repo).",
 	}
-	err := survey.AskOne(prompt, &accessToken, makeAccessTokenCheck(serviceRepo))
+	// err := survey.AskOne(prompt, &accessToken, makeAccessTokenCheck(serviceRepo))
```
wtam2018, line 199: Remove ^^^

pkg/cmd/ui/ui.go
```diff
@@ -133,29 +160,47 @@ func enterSealedSecretService() string {
...
+	err := survey.AskOne(prompt, &accessToken, survey.Required)
 	handleError(err)
-	return accessToken
+	// err = ValidateAccessToken(accessToken, serviceRepo)
```
wtam2018, line 202: Remove ***

pkg/pipelines/git/repository_test.go
```diff
@@ -151,3 +152,31 @@ func TestCreateWebHook(t *testing.T) {
...
+		desc     string
+		argument string
+		want     string
+	}{
+		{"Standard url",
+			"https://github.com/username/repo.git",
+			`username/repo`},
+		{"Standard url",
+			"https://github.com/username/repo/.git",
+			`username/repo`},
```
wtam2018, line 167: How about adding a test case for the case you are fixing?
"https://github.com/username/repo/"

wtam2018 (changes requested):
I don't understand why there are much changes to accept an repo URL like "https://github.com/username/repo/.   I must be missing something.

PR #110, opened 2020-12-10
Add app labels to applications

bigkevmcd:
**What type of PR is this?**
> /kind enhancement

It adds two labels to objects that are created

"app.openshift.io/vcs-source" to all objects that are created, with a value that is the "fullname" of the GitOps Repo.

"app.kubernetes.io/name" is added to all ArgoCD Application objects, with the source application that they come from.

Fixes #?

### DIFF
--- a/pkg/pipelines/argocd/argocd.go
+++ b/pkg/pipelines/argocd/argocd.go
@@ -16,6 +16,11 @@ import (
 	res "github.com/redhat-developer/kam/pkg/pipelines/resources"
 )
 
+const (
+	vcsSourceLabel = "app.openshift.io/vcs-source"
+	appLabel       = "app.kubernetes.io/name"
+)
+
 var (
 	applicationTypeMeta = meta.TypeMeta(
 		"Application",
@@ -105,7 +110,7 @@ func (b *argocdBuilder) Application(env *config.Environment, app *config.Applica
 	argoFiles := res.Resources{}
 	filename := filepath.Join(basePath, env.Name+"-"+app.Name+"-app.yaml")
 
-	argoFiles[filename] = makeApplication(env.Name+"-"+app.Name, b.argoNS,
+	argoFiles[filename] = makeApplication(app, env.Name+"-"+app.Name, b.argoNS,
 		defaultProject,
 		env.Name,
 		clusterForEnv(env),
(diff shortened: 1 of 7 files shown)


pkg/pipelines/argocd/argocd.go
```diff
@@ -16,6 +16,11 @@ import (
 	res "github.com/redhat-developer/kam/pkg/pipelines/resources"
 )
 
+const (
+	vcsSourceLabel = "app.openshift.io/vcs-source"
```
wtam2018, line 20: This const is not used.

wtam2018: @bigkevmcd please check ^^^

pkg/pipelines/argocd/argocd.go
```diff
@@ -203,10 +209,18 @@ func argoCDResource(ns string) (*argov1.ArgoCD, error) {
 	}, nil
 }
 
-func makeApplication(appName, argoNS, project, ns, server string, source *argoappv1.ApplicationSource) *argoappv1.Application {
+func makeApplication(app *config.Application, appName, argoNS, project, ns, server string, source *argoappv1.ApplicationSource) *argoappv1.Application {
+	options := []meta.ObjectMetaOpt{}
+	if app != nil {
+		options = append(options, meta.AddLabels(map[string]string{
+			appLabel: app.Name,
```
keithchong, line 216: Just to confirm if I read this correctly, this will generate a label with this format:

`app.kubernetes.io/name=myAppName`

where from the code, `app.Name` is, as an example, has a value of `myAppName`

Is there any advantage if we use a KAM-specific or OpenShift-specific label for those apps generated by kam?

eg. `app.openshift.io/name` perhaps?

If we are linking to ArgoCD from dev console, do we need to filter based on only these types of apps?

bigkevmcd: `app.kubernetes.io/name` is one of the recommended names, but see here https://github.com/openshift/odo/issues/599 and this https://github.com/gorkem/app-labels/blob/master/labels-annotation-for-openshift.adoc for some more detail WRT this and OpenShift.

Essentially, this is simplifying the number of labels needed, the `app.kubernetes.io/name` label is used by the Topology view to group components too, and adding `app.kubernetes.io/name` and possibly `app.openshift.io/name` would mean that we'd presumably have two labels with the same value?

keithchong: Ok, makes sense.

PR #113, opened 2020-12-11
Trailing slash correction during bootstrap service add.

gaganhegde:
**What type of PR is this?**
> /kind bug

The URL in bootstrap and service add command do not accept URL of the form https://github.com/org/repo-name/ although it is a valid url. The changes are to accept it with unit tests.

Is an edge case does not require documentation changes.

https://issues.redhat.com/browse/GITOPS-492

### DIFF
--- a/pkg/cmd/utility/utility.go
+++ b/pkg/cmd/utility/utility.go
@@ -22,8 +22,12 @@ func AddGitSuffixIfNecessary(url string) string {
 	if url == "" || strings.HasSuffix(strings.ToLower(url), ".git") {
 		return url
 	}
-	notifyUsers(url)
-	return strings.TrimSuffix(url, "/") + ".git"
+	trimmed := strings.TrimSuffix(url, "/")
+	if trimmed != url {
+		log.Italicf(`Trimmed "/" from the end of %q"`, url)
+	}
+	log.Italicf("Adding .git to %s", trimmed)
+	return trimmed + ".git"
 }
 
 // RemoveEmptyStrings returns a slice with all the empty strings removed from the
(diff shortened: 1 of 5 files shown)


wtam2018:
This is literaly a one-liner fix.
image

pkg/cmd/utility/utility_test.go
```diff
@@ -24,11 +24,14 @@ func TestAddGitSuffix(t *testing.T) {
 		{"suffix for empty string", "", ""},
 		{"suffix already present", "https://github.com/test/org.git", "https://github.com/test/org.git"},
 		{"suffix with a different case", "https://github.com/test/org.GIT", "https://github.com/test/org.GIT"},
+		{"trailing slash absent", "https://github.com/test/org.git", "https://github.com/test/org.git"},
+		{"trailing slash present", "https://github.com/test/org/.git", "https://github.com/test/org.git"},
```
wtam2018, line 28: https://github.com/test/org/.git is a bad URL and should be rejected.

pkg/cmd/utility/utility.go
```diff
@@ -22,8 +22,8 @@ func AddGitSuffixIfNecessary(url string) string {
 	if url == "" || strings.HasSuffix(strings.ToLower(url), ".git") {
 		return url
 	}
-	log.Italicf("Adding .git to %s", url)
-	return url + ".git"
+	notifyUsers(url)
```
bigkevmcd, line 25: I think we can improve this,

Let's merge the behaviour of `notifyUsers` back into this, but do something like...

```Go
trimmed := strings.TrimSuffix(url, "/")

if trimmed != url {
  log.Italicf(`Trimmed "/" from the end of %q", url)
}

return trimmed + ".git"

bigkevmcd (approved):

PR #129, opened 2020-12-28
Improve sealedsecret namespace

haf-tech:
**What type of PR is this?**
/kind enhancement

In the bootstrap use case does the checkDependency logic does not take into account if the user provide custom values for SealedSecrets (service-name, namespace). This results in wrong info in the spinner and let the impression that something is wrong during bootstrap.

This enhancement is necessary, because the SealedSecrets Operator in OpenShift is installed in other namespace (not `cicd`)

No docu update necessary

n/a
Fixes #?

* Test boostrap with the parameters `--sealed-secrets-ns` and `--sealed-secrets-svc`

```
dist/kam_darwin_amd64 bootstrap \
  --service-repo-url https://github.com/k8s-universe/tekton-101 \
  --gitops-repo-url https://github.com/k8s-universe/tekton-101-gitops.git \
  --image-repo image-registry-openshift-image-registry.apps.cluster-608e.sandbox1691.opentlc.com/demo-tekton-gitops \
  --git-host-access-token 8234db.... \
  --output . \
  --push-to-git=false \
  --sealed-secrets-ns=sealed-secrets \
  --sealed-secrets-svc=sealed-secrets-controller


Checking dependencies

 ✓  Checking if Sealed Secrets is installed with the default configuration [160ms]
 ✓  Checking if ArgoCD Operator is installed with the default configuration [343ms]
 ✓  Checking if OpenShift Pipelines Operator is installed with the default configuration [373ms]
Adding .git to https://github.com/k8s-universe/tekton-101

Completing Bootstrap process

 ✓  Authentication tokens encrypted in secrets
 ✓  Pipelines tracker has been configured
 ✓  OpenShift Pipelines resources created
 ✓  Openshift Route for EventListener created
 ✓  Created dev, stage and CICD environments
 ✓  Bootstrapped OpenShift resources successfully

 Next Steps:
 Please refer to https://github.com/redhat-developer/kam/tree/master/docs to get started.
```

Verify also the updated test cases.

### DIFF
--- a/pkg/cmd/bootstrap.go
+++ b/pkg/cmd/bootstrap.go
@@ -225,10 +225,29 @@ func checkBootstrapDependencies(io *BootstrapParameters, client *utility.Client,
 
 	spinner.Start("Checking if Sealed Secrets is installed with the default configuration", false)
 	if err := client.CheckIfSealedSecretsExists(defaultSealedSecretsServiceName); err != nil {
-		warnIfNotFound(spinner, "Please install Sealed Secrets Operator from OperatorHub", err)
-		if !apierrors.IsNotFound(err) {
-			return fmt.Errorf("failed to check for Sealed Secrets Operator: %w", err)
+
+		// in case Sealed Secret namespace/service name are provided, try them too
+		if (len(io.BootstrapOptions.SealedSecretsService.Namespace) > 0 && io.BootstrapOptions.SealedSecretsService.Namespace != defaultSealedSecretsServiceName.Namespace) ||
+			(len(io.BootstrapOptions.SealedSecretsService.Name) > 0 && (io.BootstrapOptions.SealedSecretsService.Name != defaultSealedSecretsServiceName.Name)) {
+
+			if err := client.CheckIfSealedSecretsExists(io.BootstrapOptions.SealedSecretsService); err != nil {
+				warnIfNotFound(spinner, "Provided Sealed Secrets namespace/name are not valid. Please verify", err)
+				if !apierrors.IsNotFound(err) {
+					return fmt.Errorf("failed to check for Sealed Secrets Operator: %w", err)
+				}
+			} else {
+				io.SealedSecretsService = io.BootstrapOptions.SealedSecretsService
+			}
+
+		} else {
+			// no dedicated name/namespace given, and also default one not found
+
+			warnIfNotFound(spinner, "Please install Sealed Secrets Operator from OperatorHub", err)
+			if !apierrors.IsNotFound(err) {
+				return fmt.Errorf("failed to check for Sealed Secrets Operator: %w", err)
+			}
 		}
+
 		// We do not add Sealed Secret Operator to missingDeps since we this dependency can be resolved
 		// by optional flags or interactive user inputs.
 	} else {
--- a/pkg/cmd/bootstrap.go
+++ b/pkg/cmd/bootstrap.go
@@ -223,35 +223,31 @@ func checkBootstrapDependencies(io *BootstrapParameters, client *utility.Client,
 	missingDeps := []string{}
 	log.Progressf("\nChecking dependencies\n")
 
-	spinner.Start("Checking if Sealed Secrets is installed with the default configuration", false)
-	if err := client.CheckIfSealedSecretsExists(defaultSealedSecretsServiceName); err != nil {
+	// in case custom Sealed Secrets namespace/service name are provided, try them first
+	// We do not add Sealed Secret Operator to missingDeps since we this dependency can be resolved
+	// by optional flags or interactive user inputs.
+	if (io.BootstrapOptions.SealedSecretsService.Namespace != "" && io.BootstrapOptions.SealedSecretsService.Namespace != defaultSealedSecretsServiceName.Namespace) ||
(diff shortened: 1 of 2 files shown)


shubhamagarwal19:
i think if a user is providing a custom name and namespace as part of bootstrap options we should only check for that name and namespace and not the default values. Default values should only be used if those flags aren't provided by the user and is empty. 
This could cause confusion as user might think the service is installed in the namespace provided, while it may have used the default values to validate the dependency.

haf-tech:
@shubhamagarwal19 yes agree, but I do not want to break the existing logic, which always check first for default, then the custom config. I do not know the history of this decision.

pkg/cmd/bootstrap.go
```diff
@@ -225,10 +225,29 @@ func checkBootstrapDependencies(io *BootstrapParameters, client *utility.Client,
 
 	spinner.Start("Checking if Sealed Secrets is installed with the default configuration", false)
 	if err := client.CheckIfSealedSecretsExists(defaultSealedSecretsServiceName); err != nil {
-		warnIfNotFound(spinner, "Please install Sealed Secrets Operator from OperatorHub", err)
-		if !apierrors.IsNotFound(err) {
-			return fmt.Errorf("failed to check for Sealed Secrets Operator: %w", err)
+
+		// in case Sealed Secret namespace/service name are provided, try them too
+		if (len(io.BootstrapOptions.SealedSecretsService.Namespace) > 0 && io.BootstrapOptions.SealedSecretsService.Namespace != defaultSealedSecretsServiceName.Namespace) ||
```
bigkevmcd, line 230: I don't like to use string length comparisons to check if a string is empty, it compiles to the same as testing for `""` and it's a lot more obvious what is actually being compared.

pkg/cmd/bootstrap_test.go
```diff
@@ -34,6 +35,11 @@ const (
 	serviceURL = "https://github.com/org/app"
 )
 
+const (
+	customSealedSecretsNS         = "sealed-secrets"
+	customSealedSecretsController = "cutom-sealed-secrets-controller"
```
bigkevmcd, line 40: I think this is a typo `s/cutom/custom/` ?

bigkevmcd (changes requested):
@haf-tech thanks for the contribution, I agree with @shubhamagarwal19 that we should check for the **default** _or_ the user provided value.

pkg/cmd/bootstrap.go
haf-tech: yes, will change to `!= ""`

haf-tech:
@bigkevmcd 
> @haf-tech thanks for the contribution, I agree with @shubhamagarwal19 that we should check for the **default** _or_ the user provided value.

I provided a new commit with the fixes.
But I did not touched the default-logic. Let's agree about the target approach
1. if *user* defined sealed secret information are given, use this
2. if no *user* defined info are given, use the default one

open question is
should we fallback to 2, if 1 fails?

bigkevmcd:
> But I did not touched the default-logic. Let's agree about the target approach
> 
> 1. if _user_ defined sealed secret information are given, use this
> 2. if no _user_ defined info are given, use the default one
> 
> open question is
> should we fallback to 2, if 1 fails?

I'd say no, it's probably confusing if you specify a specific namespace/name combination, but with a typo, and it still succeeds, because we found the default one?

What do you think?

haf-tech:
> > But I did not touched the default-logic. Let's agree about the target approach
> > 
> > 1. if _user_ defined sealed secret information are given, use this
> > 2. if no _user_ defined info are given, use the default one
> > 
> > open question is
> > should we fallback to 2, if 1 fails?
> 
> I'd say no, it's probably confusing if you specify a specific namespace/name combination, but with a typo, and it still succeeds, because we found the default one?
> 
> What do you think?

yes, agree. this is not transparent and confusing
then I would adjust the logic to meet this approach - enhancing this PR, are you fine with it?

bigkevmcd:
> > > But I did not touched the default-logic. Let's agree about the target approach
> > > 
> > > 1. if _user_ defined sealed secret information are given, use this
> > > 2. if no _user_ defined info are given, use the default one
> > > 
> > > open question is
> > > should we fallback to 2, if 1 fails?
> > 
> > 
> > I'd say no, it's probably confusing if you specify a specific namespace/name combination, but with a typo, and it still succeeds, because we found the default one?
> > What do you think?
> 
> yes, agree. this is not transparent and confusing
> then I would adjust the logic to meet this approach - enhancing this PR, are you fine with it?

Yes, I'd be happy with this, thank you.

bigkevmcd:
@haf-tech just checking in in case you're expecting something from me, I'm happy with the behaviour outlined above.

haf-tech:
> @haf-tech just checking in in case you're expecting something from me, I'm happy with the behaviour outlined above.

not yet, last week on vacation :-) Will do the change to the weekend and then I would request a review again from your side :-)

haf-tech:
@bigkevmcd your feedback is now highly welcome :-) 
The change is imo very simple and I do not see any similar parts in the code base which needs adjustments.

bigkevmcd (approved):
This looks good, thank you!

PR #139, opened 2021-01-06
Add a dockerfile to create an image to serve kam binary

sbose78:
Can we build an image and serve `kam` from a container in OpenShift ?


amitkrout:
Platform issue
/retest

bigkevmcd (approved):
This looks reasonable to me

Issue #147, opened 2021-01-15
kam fails to write resources to output folder 'B' after user is prompted to not overwrite the contents of folder 'A'


keithchong:
**Describe the bug**
kam prompts the user to provide the output folder of the generated resource files.  If that folder already has the generated files from a previous run (includes pipelines.yaml), then the user is prompted to overwrite or change the output path.  If the user chooses another folder (an empty folder for example), then the resources are not generated and no error is reported.

**To Reproduce**
Make sure you already have kam-generated code somewhere in your filesystem. Let's refer to this folder as `A`.
1. Run `kam bootstrap` and follow all questions until you are asked to provide the output path
2. Provide the path for folder `A`
3. You will be prompted to overwrite
4. Select No
5. You will be prompted to provide another path
6. Specify a new path to an empty folder, call this, path `B`
7. Proceed with the rest of the prompts to end.  kam reports everything is ok.
==> Check path `B` and it will still be empty

**Expected behavior**
Path `B` should have the generated files

**Screenshots**
Here is a sample output:
```

kam bootstrap
  
Checking dependencies

 ✓  Checking if Sealed Secrets is installed with the default configuration [347ms]
 ✓  Checking if ArgoCD Operator is installed with the default configuration [567ms]
 ✓  Checking if OpenShift Pipelines Operator is installed with the default configuration [654ms]
  
Starting interactive prompt

? Provide the URL for your GitOps repository https://github.com/keithchong/gitops9.git
? Select type of image registry External Registry
? Image registry of the form <registry>/<username>/<image name> which is used to push newly built images. quay.io/keithchong/taxi
? Path to config.json which authenticates image pushes to the desired image registry /Users/kchong/Downloads/keithchong-robot-auth.json
? Provide a secret (minimum 16 characters) that we can use to authenticate incoming hooks from your Git hosting service for the Service repository. (if not provided, it will be auto-generated) 
? Provide the URL for your Service repository e.g. https://github.com/organisation/service.git https://github.com/keithchong/taxi.git
? Is this repository a private repository? no
? Provide a secret (minimum 16 characters) that we can use to authenticate incoming hooks from your Git hosting service for the Service repository. (if not provided, it will be auto-generated) 
? Do you want to enable commit-status-tracker? yes
? Please provide a token used to authenticate requests to "https://github.com/keithchong/taxi.git" [? for help] ****************************************
? Add a prefix to the environment names(dev, stage, cicd etc.) to distinguish and identify individual environments? 
[...]

PR #148, opened 2021-01-15
Do not merge: Smart default for bootstrap options except for point 5 of the epic; Fix 147

keithchong:

**What type of PR is this?**
This is for the smart defaulting of the output folder for `kam bootstrap`.  The user will not be prompted to provide an output folder if the default current folder is used or if the `--output string` is used.  This also includes a bug fix for https://github.com/redhat-developer/kam/issues/147 , since testing of the feature uncovered the bug.

/kind bug - Fixes 147
/kind enhancement - See GitOps story 563

See
GitOps story 563
https://github.com/redhat-developer/kam/issues/147

Fixes https://github.com/redhat-developer/kam/issues/147

Run `kam bootstrap` and see if you get prompted for the output folder.  Check that the resources are written to the current folder.

### DIFF
--- a/pkg/cmd/bootstrap.go
+++ b/pkg/cmd/bootstrap.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"errors"
 	"fmt"
+	"github.com/redhat-developer/kam/pkg/pipelines/imagerepo"
 	"net/url"
 	"os"
 	"strings"
@@ -71,7 +72,8 @@ var (
 // BootstrapParameters encapsulates the parameters for the kam pipelines init command.
 type BootstrapParameters struct {
 	*pipelines.BootstrapOptions
-	PushToGit bool // records whether or not the repository should be pushed to git.
+	PushToGit   bool // records whether or not the repository should be pushed to git.
+	Interactive bool
 }
 
 type status interface {
@@ -108,9 +110,10 @@ func (io *BootstrapParameters) Complete(name string, cmd *cobra.Command, args []
 		return err
 	}
 
-	if cmd.Flags().NFlag() == 0 {
-		return initiateInteractiveMode(io, client)
+	if cmd.Flags().NFlag() == 0 || cmd.Flag("interactive").Changed {
+		return initiateInteractiveMode(io, client, cmd)
 	}
+
 	addGitURLSuffixIfNecessary(io)
 	return nonInteractiveMode(io, client)
 }
@@ -122,7 +125,7 @@ func addGitURLSuffixIfNecessary(io *BootstrapParameters) {
 
 // nonInteractiveMode gets triggered if a flag is passed, checks for mandatory flags.
 func nonInteractiveMode(io *BootstrapParameters, client *utility.Client) error {
-	mandatoryFlags := map[string]string{serviceRepoURLFlag: io.ServiceRepoURL, gitopsRepoURLFlag: io.GitOpsRepoURL, imageRepoFlag: io.ImageRepo}
+	mandatoryFlags := map[string]string{serviceRepoURLFlag: io.ServiceRepoURL, gitopsRepoURLFlag: io.GitOpsRepoURL}
 	if err := checkMandatoryFlags(mandatoryFlags); err != nil {
 		return err
 	}
@@ -135,7 +138,7 @@ func nonInteractiveMode(io *BootstrapParameters, client *utility.Client) error {
 
 func checkMandatoryFlags(flags map[string]string) error {
 	missingFlags := []string{}
-	mandatoryFlags := []string{serviceRepoURLFlag, gitopsRepoURLFlag, imageRepoFlag}
+	mandatoryFlags := []string{serviceRepoURLFlag, gitopsRepoURLFlag}
 	for _, flag := range mandatoryFlags {
 		if flags[flag] == "" {
 			missingFlags = append(missingFlags, fmt.Sprintf("%q", flag))
@@ -152,13 +155,18 @@ func missingFlagErr(flags []string) error {
 }
 
 // initiateInteractiveMode starts the interactive mode impplementation if no flags are passed.
-func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client) error {
+func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client, cmd *cobra.Command) error {
 	log.Progressf("\nStarting interactive prompt\n")
+	// Prompt if user wants to use all default values and only be prompted with required or other necessary questions
+	promptForAll := !ui.UseDefaultValues()
 	// ask for sealed secrets only when default is absent
 	if client.CheckIfSealedSecretsExists(defaultSealedSecretsServiceName) != nil {
 		io.SealedSecretsService.Namespace = ui.EnterSealedSecretService(&io.SealedSecretsService)
 	}
-	io.GitOpsRepoURL = utility.AddGitSuffixIfNecessary(ui.EnterGitRepo())
+	if io.GitOpsRepoURL == "" {
+		io.GitOpsRepoURL = ui.EnterGitRepo()
+	}
+	io.GitOpsRepoURL = utility.AddGitSuffixIfNecessary(io.GitOpsRepoURL)
 	if !isKnownDriver(io.GitOpsRepoURL) {
 		io.PrivateRepoDriver = ui.SelectPrivateRepoDriver()
 		host, err := accesstoken.HostFromURL(io.GitOpsRepoURL)
@@ -168,32 +176,80 @@ func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client) er
 		identifier := factory.NewDriverIdentifier(factory.Mapping(host, io.PrivateRepoDriver))
 		factory.DefaultIdentifier = identifier
 	}
-	if ui.UseInternalRegistry() {
-		io.ImageRepo = ui.EnterImageRepoInternalRegistry()
-	} else {
-		io.ImageRepo = ui.EnterImageRepoExternalRepository()
-		io.DockerConfigJSONFilename = ui.EnterDockercfg()
+	if io.ImageRepo != "" {
+		isInternalRegistry, _, err := imagerepo.ValidateImageRepo(io.ImageRepo)
+		if !isInternalRegistry {
+			if !cmd.Flag("dockercfgjson").Changed {
+				log.Progressf("The supplied image repository has been detected as an external repository.")
+				io.DockerConfigJSONFilename = ui.EnterDockercfg()
+			}
+		}
+		if err != nil {
+			return err
+		}
+	} else if promptForAll {
+		if ui.UseInternalRegistry() {
+			io.ImageRepo = ui.EnterImageRepoInternalRegistry()
+		} else {
+			io.ImageRepo = ui.EnterImageRepoExternalRepository()
+			io.DockerConfigJSONFilename = ui.EnterDockercfg()
+		}
 	}
-	io.GitOpsWebhookSecret = ui.EnterGitWebhookSecret(io.GitOpsRepoURL)
-	io.ServiceRepoURL = utility.AddGitSuffixIfNecessary(ui.EnterServiceRepoURL())
-	io.ServiceWebhookSecret = ui.EnterGitWebhookSecret(io.ServiceRepoURL)
+	if promptForAll {
+		io.GitOpsWebhookSecret = ui.EnterGitWebhookSecret(io.GitOpsRepoURL)
+	}
+	if io.ServiceRepoURL == "" {
+		io.ServiceRepoURL = ui.EnterServiceRepoURL()
+	}
+	io.ServiceRepoURL = utility.AddGitSuffixIfNecessary(io.ServiceRepoURL)
+	if promptForAll {
+		io.ServiceWebhookSecret = ui.EnterGitWebhookSecret(io.ServiceRepoURL)
+	}
 	secret, err := accesstoken.GetAccessToken(io.ServiceRepoURL)
 	if err != nil && err != keyring.ErrNotFound {
 		return err
 	}
-	if secret == "" {
-		io.GitHostAccessToken = ui.EnterGitHostAccessToken(io.ServiceRepoURL)
-		io.SaveTokenKeyRing = ui.UseKeyringRingSvc()
+	if secret == "" { // We must prompt for the token
+		if io.GitHostAccessToken == "" {
+			io.GitHostAccessToken = ui.EnterGitHostAccessToken(io.ServiceRepoURL)
+		}
+		if !cmd.Flag("save-token-keyring").Changed {
+			io.SaveTokenKeyRing = ui.UseKeyringRingSvc()
+		}
 		setAccessToken(io)
 	} else {
 		io.GitHostAccessToken = secret
 	}
-	io.CommitStatusTracker = ui.SetupCommitStatusTracker()
-	io.PushToGit = ui.SelectOptionPushToGit()
-	io.Prefix = ui.EnterPrefix()
-	io.OutputPath = ui.EnterOutputPath()
+	if !cmd.Flag("commit-status-tracker").Changed && promptForAll {
+		io.CommitStatusTracker = ui.SetupCommitStatusTracker()
+	}
+	if !cmd.Flag("push-to-git").Changed && promptForAll {
+		io.PushToGit = ui.SelectOptionPushToGit()
+	}
+	if io.Prefix == "" && promptForAll {
+		io.Prefix = ui.EnterPrefix()
+	}
+	outputPathOverridden := cmd.Flag("output").Changed
+	if !outputPathOverridden {
+		// Override the default path to be ./{gitops repo name}
+		repoName, err := repoFromURL(io.GitOpsRepoURL)
+		if err != nil {
+			repoName = "gitops"
+		}
+		io.OutputPath = "./" + repoName
+	}
+	io.OutputPath = ui.VerifyOutputPath(io.OutputPath, io.Overwrite, outputPathOverridden, promptForAll)
 	io.Overwrite = true
 	return nil
+}
+
+func repoFromURL(raw string) (string, error) {
+	u, err := url.Parse(raw)
+	if err != nil {
+		return "", err
+	}
+	parts := strings.Split(u.Path, "/")
+	return strings.TrimSuffix(parts[len(parts)-1], ".git"), nil
 }
 
 func setAccessToken(io *BootstrapParameters) error {
@@ -353,7 +409,7 @@ func NewCmdBootstrap(name, fullName string) *cobra.Command {
 	}
 	bootstrapCmd.Flags().StringVar(&o.GitOpsRepoURL, "gitops-repo-url", "", "Provide the URL for your GitOps repository e.g. https://github.com/organisation/repository.git")
 	bootstrapCmd.Flags().StringVar(&o.GitOpsWebhookSecret, "gitops-webhook-secret", "", "Provide a secret that we can use to authenticate incoming hooks from your Git hosting service for the GitOps repository. (if not provided, it will be auto-generated)")
-	bootstrapCmd.Flags().StringVar(&o.OutputPath, "output", ".", "Path to write GitOps resources")
+	bootstrapCmd.Flags().StringVar(&o.OutputPath, "output", "./gitops", "Path to write GitOps resources")
 	bootstrapCmd.Flags().StringVarP(&o.Prefix, "prefix", "p", "", "Add a prefix to the environment names(Dev, stage,prod,cicd etc.) to distinguish and identify individual environments")
 	bootstrapCmd.Flags().StringVar(&o.DockerConfigJSONFilename, "dockercfgjson", "~/.docker/config.json", "Filepath to config.json which authenticates the image push to the desired image registry ")
 	bootstrapCmd.Flags().StringVar(&o.ImageRepo, "image-repo", "", "Image repository of the form <registry>/<username>/<repository> or <project>/<app> which is used to push newly built images")
@@ -367,6 +423,7 @@ func NewCmdBootstrap(name, fullName string) *cobra.Command {
 	bootstrapCmd.Flags().StringVar(&o.PrivateRepoDriver, "private-repo-driver", "", "If your Git repositories are on a custom domain, please indicate which driver to use github or gitlab")
 	bootstrapCmd.Flags().BoolVar(&o.CommitStatusTracker, "commit-status-tracker", true, "Enable or disable the commit-status-tracker which reports the success/failure of your pipelineruns to GitHub/GitLab")
 	bootstrapCmd.Flags().BoolVar(&o.PushToGit, "push-to-git", false, "If true, automatically creates and populates the gitops-repo-url with the generated resources")
+	bootstrapCmd.Flags().BoolVar(&o.Interactive, "interactive", false, "If true, enable prompting for most options if not already specified on the command line")
 	return bootstrapCmd
 }
 
--- a/pkg/cmd/ui/ui.go
+++ b/pkg/cmd/ui/ui.go
@@ -2,12 +2,11 @@ package ui
 
 import (
 	"fmt"
+	"github.com/redhat-developer/kam/pkg/pipelines/ioutils"
+	"gopkg.in/AlecAivazis/survey.v1"
 	"net/url"
 	"path/filepath"
 
-	"gopkg.in/AlecAivazis/survey.v1"
-
-	"github.com/redhat-developer/kam/pkg/pipelines/ioutils"
 	"k8s.io/apimachinery/pkg/types"
 )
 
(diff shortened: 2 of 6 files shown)


pkg/cmd/ui/ui.go
```diff
@@ -86,18 +85,24 @@ func EnterImageRepoExternalRepository() string {
 }
 
 // EnterOutputPath allows the user to specify the path where the gitops configuration must reside locally in a UI prompt.
-func EnterOutputPath() string {
+func EnterOutputPath(originalPath string) string {
```
bigkevmcd, line 88: I think we can simplify the logic in this function significantly.

pkg/cmd/ui/ui.go
```diff
@@ -86,18 +85,24 @@ func EnterImageRepoExternalRepository() string {
...
 	var outputPath string
-	prompt := &survey.Input{
-		Message: "Provide a path to write GitOps resources?",
-		Help:    "This is the path where the GitOps repository configuration is stored locally before you push it to the repository GitopsRepoURL",
-		Default: ".",
+	var err error
```
bigkevmcd, line 90: You don't need to declare the error.

pkg/cmd/ui/ui.go
```diff
@@ -86,18 +85,24 @@ func EnterImageRepoExternalRepository() string {
...
+	if originalPath == "" {
+		prompt := &survey.Input{
+			Message: "Provide a path to write GitOps resources?",
+			Help:    "This is the path where the GitOps repository configuration is stored locally before you push it to the repository GitopsRepoURL",
+			Default: ".",
+		}
+		err = survey.AskOne(prompt, &outputPath, nil)
```
bigkevmcd, line 97: You could simplify this to `handleError(survey.AskOne(prompt, &outputPath, nil))`

pkg/cmd/ui/ui.go
```diff
@@ -86,18 +85,24 @@ func EnterImageRepoExternalRepository() string {
...
+		handleError(err)
 	}
-
-	err := survey.AskOne(prompt, &outputPath, nil)
 	exists, filePathError := ioutils.IsExisting(ioutils.NewFilesystem(), filepath.Join(outputPath, "pipelines.yaml"))
 	if exists {
```
bigkevmcd, line 101: This can become `exists, err := ...`

And normally, you'd return if the error was not nil.

pkg/cmd/ui/ui.go
```diff
@@ -223,7 +228,7 @@ func SelectOptionOverwrite(path string) string {
 		Options: []string{"yes", "no"},
 		Default: "no",
 	}
-	err := survey.AskOne(prompt, &overwrite, makeOverWriteValidator(path))
+	err := survey.AskOne(prompt, &overwrite, nil)
```
bigkevmcd, line 231: With the changes above, you can simplify this code a fair bit too.

bigkevmcd (changes requested):

bigkevmcd:
Looks like good start, the code can be tidied up significantly tho'.

bigkevmcd:
For brownie points...you could simplify `handleError`.

It can return early if the err is nil.

`os.Exit(1)` causes the process to terminate, so...you don't need to else it with the logging.

pkg/cmd/ui/ui.go
keithchong: Hi Kevin, could you please elaborate on this a bit more?  The command cobra command will default to "." if the flag is not specified, and if specified, it will be some non-null value that the user provides.  So the "smart default" part of this fix will pass in this value and not prompt the user to enter a path value.  It will just check if that path will be overwritten and prompt the user with the appropriate message if necessary.  If the path already contains the generated files, then this path is reset to "" (See line 104) so that the user is prompted to enter a path.

bigkevmcd: Sure, but that doesn't mean that the code can't be simplified.

keithchong: I'll remove the recursive call to the method and use a for loop instead

keithchong:
> For brownie points...you could simplify `handleError`.
> 
> It can return early if the err is nil.
> 
> `os.Exit(1)` causes the process to terminate, so...you don't need to else it with the logging.

PR updated with this change

keithchong:
Currently kam bootstrap is in interactive mode only if no options are passed to it.  However, because existing or new flags are used to override default values, that means, it should be interactive mode even if one of these optional flags are provided by the user.  For example:

`kam bootstrap --output ./genfolder`
`kam boostrap --commit-status-tracker=false`
`kam bootstrap --output ./genfolder --commit-status-tracker=false`

bigkevmcd:
> Currently kam bootstrap is in interactive mode only if no options are passed to it. However, because existing or new flags are used to override default values, that means, it should be interactive mode even if one of these optional flags are provided by the user. For example:
> 
> `kam bootstrap --output ./genfolder`
> `kam boostrap --commit-status-tracker=false`
> `kam bootstrap --output ./genfolder --commit-status-tracker=false`

Wouldn't this break scripts?

keithchong:
> > Currently kam bootstrap is in interactive mode only if no options are passed to it. However, because existing or new flags are used to override default values, that means, it should be interactive mode even if one of these optional flags are provided by the user. For example:
> > `kam bootstrap --output ./genfolder`
> > `kam boostrap --commit-status-tracker=false`
> > `kam bootstrap --output ./genfolder --commit-status-tracker=false`
> 
> Wouldn't this break scripts?

I ran `make test` and `make e2e` and the tests passed. I figure that new tests need to be created instead.

BTW, I think kam should be even smarter to prompt the user for any missing information when necessary.  For example:

```
kam bootstrap  --output ./zzout
  
Checking dependencies

 ✓  Checking if Sealed Secrets is installed with the default configuration [153ms]
 ✓  Checking if ArgoCD Operator is installed with the default configuration [372ms]
 ✓  Checking if OpenShift Pipelines Operator is installed with the default configuration [429ms]
  
Starting interactive prompt

? Provide the URL for your GitOps repository [? for help] 
```
(CTRL-C to show you will get prompted for other options including the required flags)
```
kam bootstrap  --gitops-repo-url https://github.com/keithchong/gitops.git
  
Checking dependencies

 ✓  Checking if Sealed Secrets is installed with the default configuration [149ms]
 ✓  Checking if ArgoCD Operator is installed with the default configuration [406ms]
 ✓  Checking if OpenShift Pipelines Operator is installed with the default configuration [490ms]
 ✗  required flag(s) "service-repo-url", "image-repo" not set
```

Here, this is in non-interactive mode, but we could be smart to still prompt the user for the required options.  So, if the user provide all required options ONLY, then that is when non-interactive mode should be run because the default values will be used for the other options.

keithchong:
For the changes for the smart defaults of the image registry, it might be better if it is in another PR. The flag for it allows for external image repositories (quay.io) then if the user does override it with a value, then IMHO, we still need to prompt for the config.json file if the option for it (--dockercfgjson) is not specified, and even if the default value for it is `~/docker/config.json`.  We will have to add validation of the repository value to see if it is in the external format (three segments) and then prompt accordingly.  If it's internal, then no prompt is necessary.

Here is an example of the default behaviour:

```
kam bootstrap
  
Checking dependencies

 ✓  Checking if Sealed Secrets is installed with the default configuration [252ms]
 ✓  Checking if ArgoCD Operator is installed with the default configuration [449ms]
 ✓  Checking if OpenShift Pipelines Operator is installed with the default configuration [531ms]
  
Starting interactive prompt

? Provide the URL for your GitOps repository https://github.com/keithchong/gitops9.git
? Provide the URL for your Service repository e.g. https://github.com/organisation/service.git https://github.com/keithchong/taxi.git
? Please provide a token used to authenticate requests to "https://github.com/keithchong/taxi.git" [? for help] ****************************************
? Do you wish to securely store the git-host-access-token in the keyring on your local machine? No
? Do you want to create and push the resources to your gitops repository? no
? Add a prefix to the environment names(dev, stage, cicd etc.) to distinguish and identify individual environments? pref1
? Do you want to overwrite your output path? yes
Completing Bootstrap process

The default image repository that will be used is: image-registry.openshift-image-registry.svc:5000/pref1-cicd/taxi

 ✓  Authentication tokens encrypted in secrets
 ✓  Pipelines tracker has been configured
 ✓  OpenShift Pipelines resources created
 ✓  Openshift Route for EventListener created
 ✓  Created dev, stage and CICD environments
 ✓  Bootstrapped OpenShift resources successfully

 Next Steps:
 Please refer to https://github.com/redhat-developer/kam/tree/master/docs to get started.
```

bigkevmcd (commented):
Looking at the extra rules around interactive mode, I think that `--interactive` feels like a nicer option.

pkg/cmd/bootstrap.go
```diff
@@ -152,7 +152,7 @@ func missingFlagErr(flags []string) error {
 }
 
 // initiateInteractiveMode starts the interactive mode impplementation if no flags are passed.
-func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client) error {
+func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client, cmd *cobra.Command) error {
```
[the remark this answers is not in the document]
keithchong: Yes, we will need it for handling the default and non-default values for the image repository.  We'll need to check if the flag was 'Changed'.  (I thought maybe we could log a message for our own purposes.  I'll remove the log references.)

Example 1: override image-repo with external URI.  Must prompt for config.json
```
kam bootstrap --image-repo quay.io/keithchong/taxi
  
Checking dependencies

 ✓  Checking if Sealed Secrets is installed with the default configuration [1s]
 ✓  Checking if ArgoCD Operator is installed with the default configuration [3s]
 ✓  Checking if OpenShift Pipelines Operator is installed with the default configuration [3s]
  
Starting interactive prompt

? Provide the URL for your GitOps repository https://github.com/keithchong/gitops9.git
The supplied image repository has been detected as an external repository.
? Please provide the path to config.json which authenticates image pushes to the desired image registry. ~/.docker/config.json
? Provide the URL for your Service repository e.g. https://github.com/organisation/service.git https://github.com/keithchong/taxi.git
? Please provide a token used to authenticate requests to "https://github.com/keithchong/taxi.git" [? for help] ****************************************
? Do you wish to securely store the git-host-access-token in the keyring on your local machine? No
? Do you want to create and push the resources to your gitops repository? no
? Add a prefix to the environment names(dev, stage, cicd etc.) to distinguish and identify individual environments? 
? Do you want to overwrite your output path? yes
  
Completing Bootstrap process

The provided image repository is  quay.io/keithchong/taxi
The config JSON is  ~/.docker/config.json
The updated image repository is  quay.io/keithchong/taxi

 ✓  Authentication tokens encrypted in secrets
 ✓  Pipelines tracker has been configured
 ✓  OpenShift Pipelines resources created
 ✓  Openshift Route for EventListener created
 ✓  Created dev, stage and CICD environments
 ✓  Bootstrapped OpenShift resources successfully

 Next Steps:
 Please refer to https://github.com/redhat-developer/kam/tree/master/docs to get started.
```

Example 2: override image-repo with external URI AND override the config location. You will not get prompted for image repo related information.

```
kam bootstrap --image-repo quay.io/keithchong/taxi --dockercfgjson ~/Downloads/keithchong-robot-auth.json
  
Checking dependencies

 ✓  Checking if Sealed Secrets is installed with the default configuration [840ms]
 ✓  Checking if ArgoCD Operator is installed with the default configuration [2s]
 ✓  Checking if OpenShift Pipelines Operator is installed with the default configuration [3s]
  
Starting interactive prompt

? Provide the URL for your GitOps repository https://github.com/keithchong/gitops9.git
? Provide the URL for your Service repository e.g. https://github.com/organisation/service.git https://github.com/keithchong/taxi.git
? Please provide a token used to authenticate requests to "https://github.com/keithchong/taxi.git" [? for help] ****************************************
? Do you wish to securely store the git-host-access-token in the keyring on your local machine? No
? Do you want to create and push the resources to your gitops repository? no
? Add a prefix to the environment names(dev, stage, cicd etc.) to distinguish and identify individual environments? 
? Do you want to overwrite your output path? yes
  
Completing Bootstrap process

The provided image repository is  quay.io/keithchong/taxi
The config JSON is  /Users/kchong/Downloads/keithchong-robot-auth.json
The updated image repository is  quay.io/keithchong/taxi

 ✓  Authentication tokens encrypted in secrets
 ✓  Pipelines tracker has been configured
 ✓  OpenShift Pipelines resources created
 ✓  Openshift Route for EventListener created
 ✓  Created dev, stage and CICD environments
 ✓  Bootstrapped OpenShift resources successfully

 Next Steps:
 Please refer to https://github.com/redhat-developer/kam/tree/master/docs to get started.

```

pkg/cmd/ui/ui.go
```diff
@@ -86,40 +85,30 @@ func EnterImageRepoExternalRepository() string {
...
+			doOverwrite := SelectOptionOverwrite()
+			if doOverwrite == "no" {
+				handleError(survey.AskOne(prompt, &outputPath, nil))
+			} else {
+				break
+			}
+		} else {
+			break
+		}
+		handleError(err)
```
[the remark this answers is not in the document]
keithchong: Ok, makes sense. I now need to pass the current value of the `overwrite` flag so that you won't get prompted if the user actually specified the `--overwrite` option. Look at how streamlined it is now:

```
kam bootstrap --gitops-repo-url https://github.com/keithchong/gitops9.git --service-repo-url https://github.com/keithchong/taxi.git --overwrite=true
  
Checking dependencies

 ✓  Checking if Sealed Secrets is installed with the default configuration [158ms]
 ✓  Checking if ArgoCD Operator is installed with the default configuration [348ms]
 ✓  Checking if OpenShift Pipelines Operator is installed with the default configuration [413ms]
  
Starting interactive prompt

? Please provide a token used to authenticate requests to "https://github.com/keithchong/taxi.git" [? for help] ****************************************
  
Completing Bootstrap process

 ✓  Options used:
    Service repository: https://github.com/keithchong/taxi.git
    GitOps repository: https://github.com/keithchong/gitops9.git
    Image repository: image-registry.openshift-image-registry.svc:5000/cicd/taxi
    Commit status tracker: true
    Output folder: .
    Overwrite output folder: true
  
 ✓  Authentication tokens encrypted in secrets
 ✓  Pipelines tracker has been configured
 ✓  OpenShift Pipelines resources created
 ✓  Openshift Route for EventListener created
 ✓  Created dev, stage and CICD environments
 ✓  Bootstrapped OpenShift resources successfully

 Next Steps:
 Please refer to https://github.com/redhat-developer/kam/tree/master/docs to get started.

```

keithchong:
Hi Kevin, I added the `--interactive` flag (which should be possibly called `--promptAll` or something like that.  If specified, it will prompt you for most options that you have NOT specified from the command line.  Obviously, you will get prompted for the required flags.

So for example, if simply want to override the output folder you AND accept default values, run:
`kam bootstrap --output ./somepath`

If you want to do the same but run with `--interactive` then you will get prompted for the other options, just like before.  So you can override the defaults during the interactive session.

pkg/cmd/bootstrap.go
```diff
@@ -108,8 +108,8 @@ func (io *BootstrapParameters) Complete(name string, cmd *cobra.Command, args []
 		return err
 	}
 
-	if cmd.Flags().NFlag() == 0 {
-		return initiateInteractiveMode(io, client)
+	if cmd.Flags().NFlag() == 0 || (!cmd.Flag(serviceRepoURLFlag).Changed && !cmd.Flag(gitopsRepoURLFlag).Changed && !cmd.Flag(imageRepoFlag).Changed) {
```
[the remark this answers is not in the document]
keithchong: See my other comment.  Getting rid of non-interactive mode altogether.  I have added `--interactive` which could also be called `--promptAll` or something like that.

keithchong:
@bigkevmcd , please review the latest changes.  I added a new prompt to ask the user to accept default values at the very start of the interactive prompt. If the user chooses yes, this will ensure we get the desired behaviour as defined by the acceptance criteria.  If the user chooses no, we get the same behaviour as before so users are prompted for all the questions so they do not have to know the actual flags.  Play around with it and you'll see what I mean. I think this is an acceptable solution.

pkg/cmd/bootstrap.go
```diff
@@ -3,6 +3,7 @@ package cmd
 import (
 	"errors"
 	"fmt"
+	"github.com/redhat-developer/kam/pkg/pipelines/imagerepo"
```
bigkevmcd, line 6: This is not the common way to do imports in Go.

From  https://github.com/golang/go/wiki/CodeReviewComments#imports 

>  The standard library packages are always in the first group.

I'd move this line down a bit (`goimports -w` will fix this for you automatically).

pkg/cmd/bootstrap.go
```diff
@@ -108,9 +110,10 @@ func (io *BootstrapParameters) Complete(name string, cmd *cobra.Command, args []
 		return err
 	}
 
-	if cmd.Flags().NFlag() == 0 {
-		return initiateInteractiveMode(io, client)
+	if cmd.Flags().NFlag() == 0 || cmd.Flag("interactive").Changed {
```
bigkevmcd, line 113: You could simplify this to just...

```
if cmd.Flags().NFlag() == 0 || io.Interactive {
}
```

pkg/cmd/bootstrap.go
```diff
@@ -168,34 +176,82 @@ func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client) er
...
 	}
-	if ui.UseInternalRegistry() {
-		io.ImageRepo = ui.EnterImageRepoInternalRegistry()
-	} else {
-		io.ImageRepo = ui.EnterImageRepoExternalRepository()
-		io.DockerConfigJSONFilename = ui.EnterDockercfg()
+	if cmd.Flag(imageRepoFlag).Changed {
+		isInternalRegistry, _, err := imagerepo.ValidateImageRepo(io.ImageRepo)
+		if !isInternalRegistry {
+			if !cmd.Flag("dockercfgjson").Changed {
```
bigkevmcd, line 182: This doesn't quite look right.

`dockercfgjson` has a default value, and https://godoc.org/github.com/spf13/pflag#FlagSet.Changed returns true if it's changed.

If the user doesn't change the default, they'll get prompted to enter it, even if they wanted to keep the default.

My gut feeling, is that this should be `if io.DockerConfigJSONFilename == "" {` i.e. if it's an external repo, but the auth has been cleared, prompt for a Docker config file?

pkg/cmd/bootstrap.go
```diff
@@ -168,34 +176,82 @@ func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client) er
...
+				log.Progressf("The supplied image repository has been detected as an external repository.")
+				io.DockerConfigJSONFilename = ui.EnterDockercfg()
+			}
+		}
+		if err != nil {
```
bigkevmcd, line 187: You should be checking for `err` before doing anything else here, it might be that the repo URL couldn't be parsed.

pkg/cmd/bootstrap.go
```diff
@@ -168,34 +176,82 @@ func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client) er
...
+			io.ImageRepo = ui.EnterImageRepoInternalRegistry()
+		} else {
+			io.ImageRepo = ui.EnterImageRepoExternalRepository()
+			io.DockerConfigJSONFilename = ui.EnterDockercfg()
+		}
+	}
+	if promptForAll {
+		io.GitOpsWebhookSecret = ui.EnterGitWebhookSecret(io.GitOpsRepoURL)
+	}
+	if !cmd.Flag(serviceRepoURLFlag).Changed {
```
bigkevmcd, line 201: I suspect you could just use `io.ServiceRepoURL == "" {` here instead.

pkg/cmd/bootstrap.go
```diff
@@ -168,34 +176,82 @@ func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client) er
...
-	io.ServiceWebhookSecret = ui.EnterGitWebhookSecret(io.ServiceRepoURL)
 	secret, err := accesstoken.GetAccessToken(io.ServiceRepoURL)
 	if err != nil && err != keyring.ErrNotFound {
 		return err
 	}
-	if secret == "" {
-		io.GitHostAccessToken = ui.EnterGitHostAccessToken(io.ServiceRepoURL)
-		io.SaveTokenKeyRing = ui.UseKeyringRingSvc()
+	if secret == "" { // We must prompt for the token
+		if !cmd.Flag(statustracker.CommitStatusTrackerSecret).Changed {
```
bigkevmcd, line 213: Again, given that there's no default for the CommitStatusTrackerSecret, we can simplify this to just...`if io.GitHostAccessToken == "" {`

pkg/cmd/bootstrap.go
```diff
@@ -168,34 +176,82 @@ func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client) er
...
+		io.Prefix = ui.EnterPrefix()
+	}
+	outputPathOverridden := cmd.Flag("output").Changed
+	if !outputPathOverridden {
+		// Override the default path to be ./{gitops repo name}
+		repoName, err := repoFromURL(io.GitOpsRepoURL)
+		if err != nil {
+			repoName = "gitops"
+		}
+		io.OutputPath = "./" + repoName
```
bigkevmcd, line 239: Please don't do this, use filepath.Join instead.

bigkevmcd (changes requested):
I'm not sure about the promptForAll logic, it feels like we could drop this question, also, it doesn't default to pushing to Git, which is probably unexpected.

pkg/cmd/bootstrap.go
keithchong: Odd, I'm using Goland.  Originally it was down with the other similar kam imports. Not sure how it got auto-added to the top there. 

Fixed.

pkg/cmd/bootstrap.go
keithchong: Thanks. I'll update all similar references to this form.

pkg/cmd/bootstrap.go
keithchong: Good catch. I'm missing the `&& promptForAll` here.  It should be:

`if !cmd.Flag("dockercfgjson").Changed && promptForAll {`

For not accepting defaults, you will only get prompted.  This part of the code is where we try to be smart and prompt for the config file when necessary.

In this example, the external image repo is provided on the command line, defaults are NOT accepted, and the config file is NOT overridden, so you will get prompted.
```
kam bootstrap --service-repo-url https://github.com/keithchong/taxi.git --image-repo quay.io/keithchong/taxi --interactive
  
Checking dependencies

 ✓  Checking if Sealed Secrets is installed with the default configuration [189ms]
 ✓  Checking if ArgoCD Operator is installed with the default configuration [437ms]
 ✓  Checking if OpenShift Pipelines Operator is installed with the default configuration [511ms]
  
Starting interactive prompt

? Do you want to accept all default values and be prompted only for the minimum required options? no
? Provide the URL for your GitOps repository https://github.com/keithchong/gitops.git
  The supplied image repository has been detected as an external repository.
? Provide the path to config.json which authenticates image pushes to the desired image registry [? for help] (~/.docker/config.json) 
```

And with the change `&& promptForAll` added, then when you do accept defaults, the third question above will NOT be asked.  That's the only difference.


For the last paragraph of your comment, that means the user specified the flag from the command line to be empty....or it could be an invalid location or it could be pointing to some non-config file.  I don't think it matters at this point of the interactive session. Later on, the authentication will simply fail.  So you will have to 'start over'.

So the gist of this part of the change is this.  If you specified the `--image-repo` flag from the command line, and if you are in interactive mode, we will determine if the image repo is external.  If external, we will prompt you for the location of the config file.  This prompting depends on if you have specified that flag from the command line or if you decided to accept defaults or not.  And the value of the config file location will be validated later.  It's a 'don't care' at this point, since the validation is done during `Run`

pkg/cmd/bootstrap.go
keithchong: Ok

pkg/cmd/bootstrap.go
keithchong: Hey Kevin, this should work too but it depends on the actual default value.  In general, if a flag's default value was changed (any of the values from line 383 and on), then you will have to change this part of the code too.  If we keep it this way by using `Changed`, then this part of the code does not have to change.

pkg/cmd/bootstrap.go
keithchong: Same rationale as in the previous comment.  Using "" ties the default value here when it's defined in line 383 and on.

pkg/cmd/bootstrap.go
keithchong: Ok.

keithchong:
@bigkevmcd I addressed the review comments except for what I believe are two points that I did not make changes for.

1)  I'd like to retain the use of `Changed` to keep it consistent throughout. 
2) The use of the new prompt/question to ask users if they want to use defaults.  This question is asked at the very start of interactive mode.  

If you remove this first question and we implicitly use default values for non-overridden flags, then users must override defaults by specifying the actual flags from the command line. Therefore, they must know the actual flags.

With the new question, it actually retains the previous behaviour where the user is prompted for everything if the user simply runs `kam bootstrap` with no flags (or `kam bootstrap --interactive`).  This way, the user does NOT have to know the actual flags and they can simply answer the questions to lead to a valid bootstrap configuration. This is what @wtam2018 wanted originally.

So with this approach, we get the best of both worlds.

This opens the door to adding a `--defaults` or `--useDefaults` flag but that is another story.

pkg/cmd/bootstrap.go
bigkevmcd: Right, but there's no sensible default value for the commit status tracker secret?

Nor for the service URL or the GitOps URL?

keithchong:
Closing this, since the changes have been merged via https://github.com/redhat-developer/kam/pull/158

PR #156, opened 2021-01-26
Create route for nginx

bigkevmcd:
/kind enhancement
A new Route is created during bootstrapping to make the default generated Nginx container available for testing.

* [x] Documentation has been updated.

Fixes #?

### DIFF
--- a/pkg/pipelines/eventlisteners/routes.go
+++ b/pkg/pipelines/eventlisteners/routes.go
@@ -1,4 +1,4 @@
-package routes
+package eventlisteners
 
 import (
 	"encoding/json"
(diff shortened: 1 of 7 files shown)


pkg/pipelines/eventlisteners/routes.go
```diff
@@ -0,0 +1,70 @@
+package eventlisteners
```
wtam2018, line 1: This module does not have much to do with eventlisteners.

wtam2018 (commented):
Please update documentation.

pkg/pipelines/eventlisteners/routes.go
bigkevmcd: This package is for creating an EventListener, and this creates the route to point to it.

The EventListener creates its own service, so there's no service to create a route for.

keithchong:
Test considerations:

When you bootstrap a configuration. a new file 300-route.yaml is created in (for the taxi example):

environments/{prefix-}dev/apps/app-taxi/services/taxi/base/config

When you apply the resources and check ArgoCD, a new route resource is created for the service.  Note that the service should be synchronized.

Issue #166, opened 2021-02-11
Unable to get gitlab repo name from private gitlab repo when running kam bootstrap


EsmaelTa:
We have a private gitlab repo in our company in the form of https://git.company-domain/group/subgroup/subgroup/repo.git

When I ran the kam bootstrap command I get the error around repo name. Based on the line below I belive this is due to the path in GitLab. Is there any plans for this to get fixed?

https://github.com/redhat-developer/kam/blob/e44c305488d6528684999206d1704ab4d80216a2/pkg/pipelines/git/repository.go#L88

The command I type is as follow:

`kam bootstrap  --private-repo-driver gitlab \
--sealed-secrets-ns cicd --sealed-secrets-svc sealedsecretcontroller-sealed-secrets \
--service-repo-url https://git.company-domain/group/subgroup/subgroup/taxi.git \
--gitops-repo-url https://git.company-domain/group/subgroup/subgroup/gitops.git \
--image-repo quay-registry-company-domain/admin/taxi \
--dockercfgjson user-auth.json \
--git-host-access-token <token> --save-token-keyring <token>`

The response I get is 

Checking dependencies

 `✓  Checking if Sealed Secrets is installed with custom configuration [28ms]
 ✓  Checking if ArgoCD is installed with the default configuration [122ms]
 ✓  Checking if OpenShift Pipelines Operator is installed with the default configuration [133ms]
 ✗  Please enter a valid access token for --save-token-keyring: unable to get the repo name from "https://git.company-domain/group/subgroup/subgroup/taxi.git": failed to get Git repo: /group/subgroup/subgroup/taxi.git`

PR #167, opened 2021-02-12
Improve support for GitLab project repos.

bigkevmcd:
**What type of PR is this?**

/kind bug

This fixes the behaviour when parsing GitLab's "project" repos, where a repo
can be inside a project.

The change is to assume that everything after the host is part of the repo,
thus https://gitlab.com/org/project/repo.git would be from repo
org/project/repo.

Fixes: #166 

Create a group, create a new repo in the group (cloning from an existing one that you're using).

Use the full repo URL in the service repo url e.g. https://gitlab.com/group/subgroup/repo.git

**There's a caveat around this, it doesn't currently support automatically creating a new repo within a group**

This is a bit tougher to fix, because we currently handle GitHub's weird behaviour :-(

### DIFF
--- a/pkg/pipelines/git/repository.go
+++ b/pkg/pipelines/git/repository.go
@@ -96,14 +96,14 @@ func GetRepoName(u *url.URL) (string, error) {
 			components = append(components, s)
 		}
 	}
-	if len(components) != 2 {
+	if len(components) == 1 {
 		return "", errors.New("failed to get Git repo: " + u.Path)
 	}
-	components[1] = strings.TrimSuffix(components[1], ".git")
+	components[len(components)-1] = strings.TrimSuffix(components[len(components)-1], ".git")
 	for _, s := range components {
 		if strings.Contains(s, ".") {
 			return "", errors.New("failed to get Git repo: " + u.Path)
 		}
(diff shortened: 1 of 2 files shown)


EsmaelTa:
@bigkevmcd  the line below also needs to be changed otherwise it will fail when validating 
https://github.com/redhat-developer/kam/blob/e44c305488d6528684999206d1704ab4d80216a2/pkg/cmd/bootstrap.go#L361

bigkevmcd:
> @bigkevmcd the line below also needs to be changed otherwise it will fail when validating
> https://github.com/redhat-developer/kam/blob/e44c305488d6528684999206d1704ab4d80216a2/pkg/cmd/bootstrap.go#L361

Hi @EsmaelTa thanks for that, that's validating the "gitops" repo URL which I don't yet have a fix for, or at least not a complete fix for.

Currently, this fix doesn't work for automatically creating the GitOps repo in a project/group path, because of the subtle difference between the way GitHub and GitLab's repository creation functionality works.

I'm still looking into what we can do about this.

EsmaelTa:
Thanks for your prompt feedback!

PR #170, opened 2021-02-16
Address issue with multiple namespace files being created.

bigkevmcd:
**What type of PR is this?**
/kind bug

When the internal registry is used, we generate more than one namespace file.

This causes issues in ArgoCD because there are duplicates of the same K8s object.

This ensures that the files have the same name, which means that whichever way they are created, they exist in the directory.

Fixes #?


wtam2018 (commented):
/lgtm
/approve
thanks @bigkevmcd

PR #95, opened 2020-11-25
Correct documentation w.r.t gitops repo creation.

iam-veeramalla:
**What type of PR is this?**
> /kind bug

**What does this PR do / why we need it**:
PR corrects the information provided in the docs

**Have you updated the necessary documentation?**
Yes

* [ ] Documentation has been updated.

**Which issue(s) this PR fixes**:
NA

**How to test changes / Special notes to the reviewer**:
Run `kam bootstrap` without creating a repo in github as suggested in documentation previously.

PR #197, opened 2021-03-19
[WIP] Add note about private repo workaround

reginapizza:
**What type of PR is this?**
 /kind documentation

Includes note about how to connect to a private git repo because currently there is some trouble doing so.

If necessary I can create and link to an actual doc in the repo but I figured that just linking to the docs in Argo CD was sufficient because they cover it very thoroughly.

### DIFF
--- a/docs/journey/day1/README.md
+++ b/docs/journey/day1/README.md
@@ -85,6 +85,8 @@ resources, and the resources will be pushed to your git hosting service.
 
 ## Private Repository
 
+> **_NOTE:_**  Currently there is a known issue with using private repos and the steps below may not be successful. If you wish to use a private repo, for now we recommend connecting to your repos through the Argo CD UI or CLI. Detailed instructions for this can be found in the [Argo CD docs here](https://argoproj.github.io/argo-cd/user-guide/private-repositories/). Once you have made the connection, you may manually sync your apps in Argo CD and proceed to the next section. 
+
 In case a [private repository](https://argoproj.github.io/argo-cd/user-guide/private-repositories) is used, enhance the operator generated Argo CD instance with the secret information how to connect to the git repos. 
 For example, this could be as follows:
 


reginapizza:
Test aren't passing- I tried running `go mod tidy` and `make cmd-docs` but neither changed anything. Error says "go mod state is not clean"? Not sure what that means I need to do :/ 

@wtam2018 any ideas?

varshab1210:
pr https://github.com/redhat-developer/kam/pull/198 has been merged into master. @reginapizza would you please rebase your pr

chetan-rns:
@reginapizza @wtam2018 Given that updating the CR doesn't work, should we even mention the approach in kam docs? We could directly point to the Argo CD docs page that explains configuring private repositories with CLI/UI. WDYT?

reginapizza:
@chetan-rns so you mean basically get rid of this whole section and just point to [the Argo CD docs](https://argoproj.github.io/argo-cd/user-guide/private-repositories/)? I'd be okay with that, my reasoning for not doing that is I wasn't sure if this would be getting fixed in the future and we would have to re-write it at some point.

chetan-rns:
@reginapizza I've raised a PR to fix this. But I doubt it'll go with GitOps operator v1.1 https://github.com/argoproj-labs/argocd-operator/pull/269/files

reginapizza:
@chetan-rns right, but in the meantime before that gets in shouldn't we be noting the workaround?

ishitasequeira:
@reginapizza you could mention this work around for users who have their 2FA enabled

reginapizza:
Just keeping this issue updated- 
This PR will not be needed if everything works as it's supposed to now that [PR #95 of the gitops operator](https://github.com/redhat-developer/gitops-operator/pull/95) is merged. Will retest and re-evaluate if the fix works and if this is still needed.

PR #213, opened 2021-03-29
Update default ref in the annotation URL

chetan-rns:
**What type of PR is this?**
> /kind cleanup

Once the backend [PR](https://github.com/redhat-developer/gitops-backend/pull/16) to update the default ref goes in, we should update the ref generated by kam 


amitkrout:
@chetan-rns Can you please define the test scenario in file https://github.com/redhat-developer/kam/blob/master/test/e2e/features/manual.feature if it not doable atm in any of https://github.com/redhat-developer/kam/blob/master/test/e2e/features/basic.feature or https://github.com/redhat-developer/kam/blob/master/test/e2e/features/local.feature file?

chetan-rns:
@amitkrout Since this is a small cleanup, I'm not sure if we have a separate e2e scenario for this change

PR #214, opened 2021-03-29
Adding kubectl binary in openshift CI test container

varshab1210:
**What type of PR is this?**

/kind enhancement

To get the argocd admin password we need the kubectl command to execute command ```$ kubectl get secret argocd-cluster-cluster -n openshift-gitops -ojsonpath='{.data.admin\.password}' | base64 -d```. So we need the kubectl binary available in openshift CI test container path.

Fixes #?

N/A

In the CI test container kubectl binary should be downloaded to ```bin/``` directory. For local verification you can execute the script ```install-kubectl.sh```  and check the kubectl binary should be downloaded to ```bin/``` directory

### DIFF
--- a/scripts/install-kubectl.sh
+++ b/scripts/install-kubectl.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+# Platform check
+PLATFORM=`uname -s | awk '{print tolower($0)}'`
+echo "Your platform is $PLATFORM" 
+
+# Download and install kubectl
+echo -e "\nGet kubectl binary\n"
+curl -sSL -o bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.20.5/bin/$PLATFORM/amd64/kubectl
+
+# Make the kubectl executable 
+chmod +x bin/kubectl
--- a/scripts/openshiftci-presubmit-all-tests.sh
+++ b/scripts/openshiftci-presubmit-all-tests.sh
@@ -21,6 +21,9 @@ sh $INSTALL_ARGOCD
 INSTALL_DOCKER="./scripts/install-docker-cli.sh"
 sh $INSTALL_DOCKER
 
+INSTALL_KUBECTL="./scripts/install-kubectl.sh"
+sh $INSTALL_KUBECTL
+
 export PATH="$PATH:$(pwd)/bin"
 
 # Copy kubeconfig to temporary kubeconfig file and grant


scripts/install-kubectl.sh
```diff
@@ -0,0 +1,14 @@
...
+echo "Your platform is $PLATFORM" 
+
+echo -e "\nGet kubectl binary\n"
+# Platform check for kubectl binary
+PLATFORM=`uname -s | awk '{print tolower($0)}'`
+# Download and install kubectl
+curl -sSL -o bin/kubectl https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/$PLATFORM/amd64/kubectl
+
+# Make the kubectl executable 
+chmod +x kubectl
```
jannfis, line 14: It seems you download `kubectl` to `bin/kubectl`, but the `chmod` runs on `./kubectl`.

varshab1210: > It seems you download `kubectl` to `bin/kubectl`, but the `chmod` runs on `./kubectl`.

Ahh...my bad I missed the path. Done, thanks.

amitkrout:
> To get the argocd admin password we need the kubectl command to execute command $ kubectl get secret argocd-cluster-cluster -n openshift-gitops -ojsonpath='{.data.admin\.password}' | base64 -d. So we need the kubectl binary available in openshift CI test container path.

@varshab1210 Can you please try this scenario in the local feature file. For example 
```
Scenario: Execute KAM bootstrap command with --output and --push-to-git=true flags
        When executing "kam bootstrap --service-repo-url $SERVICE_REPO_URL --gitops-repo-url $GITOPS_REPO_URL --git-host-access-token $GIT_ACCESS_TOKEN --push-to-git=true --output bootstrapresources" succeeds
        Then stderr should be empty
         When executing "oc apply -k bootstrapresources/config/argocd/" succeeds
         Then stderr should be empty
         And login "argocd-cluster-server"
```

varshab1210:
> > To get the argocd admin password we need the kubectl command to execute command $ kubectl get secret argocd-cluster-cluster -n openshift-gitops -ojsonpath='{.data.admin.password}' | base64 -d. So we need the kubectl binary available in openshift CI test container path.
> 
> @varshab1210 Can you please try this scenario in the local feature file. For example
> 
> ```
> Scenario: Execute KAM bootstrap command with --output and --push-to-git=true flags
>         When executing "kam bootstrap --service-repo-url $SERVICE_REPO_URL --gitops-repo-url $GITOPS_REPO_URL --git-host-access-token $GIT_ACCESS_TOKEN --push-to-git=true --output bootstrapresources" succeeds
>         Then stderr should be empty
>          When executing "oc apply -k bootstrapresources/config/argocd/" succeeds
>          Then stderr should be empty
>          And login "argocd-cluster-server"
> ```

Sure, I will do it in a separate pr. Thanks @amitkrout for scenario step

scripts/install-kubectl.sh
wtam2018: For testing, should we lock in a stable version of kubectl?

varshab1210: > For testing, should we lock in a stable version of kubectl?

Yes, we can do that, however one thing comes to mind that each time we need to be aware of a static kubectl version that we want to update which will increase the maintenance cost. I think we should keep a static version if its a hard requirement for certain functionalities but not sure which version to keep here. @wtam2018 can you please suggest the kubectl version
I think we we need to keep the argocd cli version static too ping @amitkrout @dewan-ahmed

amitkrout: > I think we we need to keep the argocd cli version static too ping @amitkrout @dewan-ahmed 

@jannfis Is argocd CLI release along with the argocd server ? for example OCP release has its parallel oc cli release like ocp 4.7 has oc 4.7 cli release. If yes then we can consume the static argo cli version which matches the argo server version we consume in gitops operator.

wtam2018: We can set to the current kubectl version:  v1.20.5.  Locking in a version as opposed to floating version helps to produce repeatable test results.    I would lock the argocd cli version as well.   We should know/control the tool versions that we are testing with.

varshab1210: > We can set to the current kubectl version: v1.20.5. Locking in a version as opposed to floating version helps to produce repeatable test results. I would lock the argocd cli version as well. We should know/control the tool versions that we are testing with.

Sounds good.
I will lock kubectl version in this pr and will create a separate pr to lock argocd cli version

amitkrout:
Thanks @varshab1210 for addressing all review comments. 
/approve
/lgtm

PR #215, opened 2021-03-29
Update pictures in day 1 guide to match new Argo CD instance name

reginapizza:
**What type of PR is this?**
 /kind documentation

ArgoCD instance name was changed from `argocd-cluster-cluster` to `openshift-gitops-cluster`, pictures just needed to be updated to match the text.

New pictures: 
ArgoCD_ConsoleSecrets
ArgoCD_CopyPassword


wtam2018:
@reginapizza To get secret, you don't need to switch to Developer perspective.   You can stay in Administrator perspective.   Can  you please update the screenshot and the text?

PR #219, opened 2021-04-01
Insecure option is not available in interactive mode (812)

keithchong:

**What type of PR is this?**
> /kind enhancement
We wanted to force users to specify --insecure from the command line because it is not a recommend option.  But for consistency, we should add it in interactive mode.  The message should be carefully worded to indicate it is not recommended.

Make --insecure option available in interactive mode

https://issues.redhat.com/browse/GITOPS-812

See https://issues.redhat.com/browse/GITOPS-812 for scenarios and interactive mode examples.

### DIFF
--- a/pkg/cmd/ui/ui.go
+++ b/pkg/cmd/ui/ui.go
@@ -153,6 +153,24 @@ func EnterSealedSecretService(sealedSecretService *types.NamespacedName) string 
 	return strings.TrimSpace(sealedNs)
 }
 
+// SelectInsecureSecrets, prompts the UI to ask to generate unsealed secrets or not
+func SelectInsecureSecrets(err error) bool {
+	var insecure, msg string
+	if err != nil {
+		msg = "Do you want to use 1) unsealed secrets or 2) sealed secrets and provide the details of the Sealed Secrets Operator installation?"
+	} else {
+		msg = "You are able to seal secrets. Select Sealed to continue or Unsealed to generate unsealed secrets, which is not recommended."
+	}
+	prompt := &survey.Select{
+		Message: msg,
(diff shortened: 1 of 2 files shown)


pkg/cmd/ui/ui.go
```diff
@@ -153,6 +153,24 @@ func EnterSealedSecretService(sealedSecretService *types.NamespacedName) string
 	return strings.TrimSpace(sealedNs)
 }
 
+// SelectInsecureSecrets, prompts the UI to ask to generate unsealed secrets or not
+func SelectInsecureSecrets(err error) bool {
+	var insecure, msg string
+	if err != nil {
+		msg = "Do you want to use 1) unsealed secrets or 2) sealed secrets and provide the details of the Sealed Secrets Operator installation?"
+	} else {
```
iam-veeramalla, line 161: @keithchong Do we need this ? I was just thinking if the user has already installed sealed secrets in his cluster, Why would he/she opt to not use it ?

keithchong: Hi @iam-veeramalla , in my original change for this, I did not include this.  If the dependency check succeeds, then the user should just go ahead and seal the secrets.  However, the issue here is that we need the option in interactive mode for all cases.  So, for the sake of consistency, you will still get prompted.  Another reason is that, say User A set up the sealed secrets operator and controller so that dependency check succeeded. However, User B running this command is intent on using unsealed secrets.  This provides the way to do it without using the --insecure flag on the command line.  Have a look at the Jira for the detailed scenarios.

iam-veeramalla: makes sense :) , Thanks @keithchong

chetan-rns (approved):
Thanks for the PR! Tested it locally

/lgtm

We should figure out a way to add tests for interactive mode. Will create an issue for this.

PR #221, opened 2021-04-06
fixed the error message for push-to-git if repo exists

shubhamagarwal19:
**What type of PR is this?**
/kind bug

This PR fixes the error message that is returned when `push-to-git` is enabled and the repo already exists. Earlier the error message was not informative for the user to understand the reason for failure. 

Fixes #?

Try to bootstrap a git repo that already exists. 
It should return an error message saying "repo[repo name] already exists" as the reason for failure.

PS: tried adding test cases for this, but it uses a mock provided by a vendor package and it doesn't throw an error on `CREATE repo`, if it already exists, simply overrides it.

### DIFF
--- a/pkg/pipelines/utils.go
+++ b/pkg/pipelines/utils.go
@@ -58,6 +58,13 @@ func BootstrapRepository(o *BootstrapOptions, f clientFactory, e executor) error
 	}
 	created, _, err := client.Repositories.Create(context.Background(), ri)
 	if err != nil {
+		repo := fmt.Sprintf("%s/%s", org, repoName)
+		if org == "" {
+			repo = fmt.Sprintf("%s/%s", currentUser.Login, repoName)
+		}
+		if _, resp, err := client.Repositories.Find(context.Background(), repo); err == nil && resp.Status == 200 {
+			return fmt.Errorf("failed to create repository, repo already exists")
+		}
 		return fmt.Errorf("failed to create repository %q in namespace %q: %w", repoName, org, err)
 	}
 	if err := pushRepository(o, created.CloneSSH, e); err != nil {


pkg/pipelines/utils.go
```diff
@@ -58,6 +58,13 @@ func BootstrapRepository(o *BootstrapOptions, f clientFactory, e executor) error
 	}
 	created, _, err := client.Repositories.Create(context.Background(), ri)
```
iam-veeramalla, line 59: I see that the "Create" returns (*scm.Repository, *scm.Response, error). I was thinking if the "*scm.Response" returns such error "Repo already exists"? We are ignoring "*scm.Response" in this line. Can you verify this?

iam-veeramalla: Thanks @shubhamagarwal19 .

pkg/pipelines/utils.go
```diff
@@ -58,6 +58,13 @@ func BootstrapRepository(o *BootstrapOptions, f clientFactory, e executor) error
...
 	if err != nil {
+		repo := fmt.Sprintf("%s/%s", org, repoName)
+		if org == "" {
+			repo = fmt.Sprintf("%s/%s", currentUser.Login, repoName)
+		}
+		if _, resp, err := client.Repositories.Find(context.Background(), repo); err == nil && resp.Status == 200 {
+			return fmt.Errorf("failed to create repository, repo[%s] already exists", repo)
```
chetan-rns, line 66: @shubhamagarwal19 The error message is being wrapped by the caller of `BootstrapRepository`, which results in a verbose error message

`failed to create the gitops repository: "https://github.com/chetan-rns/gitops.git": failed to create repository, repo[chetan-rns/gitops] already exists`

How about changing the message to something like

```suggestion
			return fmt.Errorf("repository already exists")
```

chetan-rns:
Looks like the mock doesn't handle this case :(
https://github.com/jenkins-x/go-scm/blob/136418105663aa37b7346f5e7d417b18a0dbc500/scm/driver/fake/repo.go#L165

pkg/pipelines/utils.go
shubhamagarwal19: done, Thanks !!

chetan-rns:
Thanks for the PR!

/lgtm
/approve

PR #224, opened 2021-04-21
[DO NOT MERGE]Test container secret path

amitkrout:
Only for CI debug purpose


amitkrout:
Purpose of knowing the secret path in the test container is solved.
/close

PR #229, opened 2021-04-26
Kam bootstrap --push-to-git flag

amitkrout:
**What type of PR is this?**
/kind enhancement

**What does this PR do / why we need it**:
Adding e2e test for --push-to-git flag

**Have you updated the necessary documentation?**

* [ ] Documentation update is required by this PR.
* [ ] Documentation has been updated.

**Which issue(s) this PR fixes**:

Fixes #127

**How to test changes / Special notes to the reviewer**:

PR #231, opened 2021-04-29
Export HOME at early stage

amitkrout:
It may fix ```Could not create directory '//.ssh': Permission denied```


amitkrout:
Merging it.
It does not break the current CI. I need this to go in ASAP for debugging https://github.com/redhat-developer/kam/pull/229

PR #233, opened 2021-05-05
Upgrade go.mod dependencies

ishitasequeira:
**What type of PR is this?**
 /kind code-refactoring

**What does this PR do / why we need it**:
RedHat Openshift Pipelines operator updated their tekton pipelines and tekton trigger components to latest versions. Due to this change, the kam bootstrap command started breaking with below error:
```
 Error from server (BadRequest): error when creating "config/cicd/base": admission webhook "webhook.triggers.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "name"
```
The issue was caused due to replacement of parameter `name` under the template section with `ref` https://github.com/tektoncd/triggers/blob/main/docs/eventlisteners.md#specifying-triggers .

As part of this update, we decided to update kam dependencies to latest versions and clean up the go.mod file for kam.
This PR includes:
- Updates to go.mod dependencies
- Fixing the code base due to api changes in updated dependencies
- Update the parameter `name` to `ref` under template in event-listeners.yaml

**Have you updated the necessary documentation?**

* [ ] Documentation update is required by this PR.
* [ ] Documentation has been updated.

**Which issue(s) this PR fixes**:

Fixes https://issues.redhat.com/browse/GITOPS-961

**How to test changes / Special notes to the reviewer**:
1. Install GitOps Operator in Openshift cluster
2. run `oc adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:openshift-gitops:openshift-gitops-argocd-application-controller`
3. oc new-project cicd
4. install Sealed Secret Operator in cicd namespace
5. Create new instance sealedsecretcontroller in cicd namespace
6. Resources of type SealedSecret stuck in the Progressing state in Argo CD(https://argo-cd.readthedocs.io/en/stable/faq/#why-are-resources-of-type-sealedsecret-stuck-in-the-progressing-state). We would need to add the below patch in Argo CD CR on the cluster. Go to openshift-gitops -> Installed Operators -> Argo CD -> openshift-gitops-> edit the yaml -> Add the below patch under the spec param ->
 ```
resourceCustomizations: | 
    bitnami.com/SealedSecret: 
      health.lua: | 
        hs = {}
        hs.status = "Healthy"
        hs.message = "Controller doesn't report resource status"
        return hs
```
7. make sure `github.com/ishitasequeira/gitops123` does not exist
8. Execute below bootstrap command
```
[...]

PR #234, opened 2021-05-06
Temporary docker container1.16

amitkrout:
**What type of PR is this?**
/kind failing-test

Fix https://github.com/redhat-developer/kam/pull/233 unit test job failure


amitkrout:
This changes has no effect on the CI run. Merging the changes into master forcefully, so that pr https://github.com/redhat-developer/kam/pull/233 will consume it.

PR #255, opened 2021-06-24
Guidance how generated secrets can be stored as SealedSecrets (#943)

keithchong:

**What type of PR is this?**

/kind enhancement

Fixes 943

See screenshots in 943.  Note that this change does not include removing the existing messages related to generating non-sealed secrets (using the --insecure flag). That should be part of 945. @ishitasequeira


pkg/cmd/bootstrap.go
```diff
@@ -432,8 +432,26 @@ func NewCmdBootstrap(name, fullName string) *cobra.Command {
 func nextSteps() {
 	log.Success("Bootstrapped OpenShift resources successfully\n\n",
 		"Next Steps:\n",
-		"Please refer to https://github.com/redhat-developer/kam/tree/master/docs to get started.",
+		"Please refer to https://github.com/redhat-developer/kam/tree/master/docs to get started.\n\n",
 	)
+	log.Info(" WARNING: Generated secrets are not encrypted.\n",
+		"  To use Bitnami Sealed Secrets:\n",
```
chetan-rns, line 438: Since we're planning to document multiple ways to manage secrets, would it be better to directly add links to the doc instead of showing steps in the command output. 
Example:
```yaml
............
............
WARNING: Generated secrets are not encrypted. For more information on secret management refer: https://github.com/redhat-developer/kam/docs/secrets

keithchong: I like your suggestion. Also, I think the external secrets set up will be more involved.

keithchong:
Hi @amitkrout , do you know why the 4.7 e2e test is failing? Looks like the 'argo-app' fails to be come synced.

wtam2018 (commented):
/lgtm
/approve

PR #260, opened 2021-07-15
The add service command introduces backslashes in kustomize.yaml (#1212)

keithchong:

**What type of PR is this?**

/kind bug

Day 2 operations will be blocked. Users probably wouldn't know immediately that to fix it, they would have to edit and change all the backward slashes in the `config\cicd\base\kustomize.yaml` file by hand.

Before running `kam service add`, the paths in the file are of the form:

`01-namespaces/cicd-environment.yaml
`

After running the command, kam incorrectly updates the file to include the OS specific file path and base folder prefix (which is why we don't see it on non-Windows platforms):

`config\cicd\base\01-namespaces\cicd-environment.yaml`

Fixes Jira 1212

I've tested the changes on Windows, following the steps in Day 2 operations. Should be in for 1.2.


pkg/pipelines/environments/environments.go
```diff
@@ -227,7 +227,7 @@ func ListFiles(fs afero.Fs, base string) (StringSet, error) {
 		if info.IsDir() {
 			return nil
 		}
-		filename := strings.TrimPrefix(path, base+"/")
+		filename := strings.TrimPrefix(filepath.ToSlash(path), filepath.ToSlash(base)+"/")
```
chetan-rns, line 230: @keithchong Thanks for the fix. We're trimming `<base>/` from the path here. Is it fine to keep the "/" hardcoded since it's direction can change with the platform?

keithchong: Yeah we need the trailing slash because the value of base that gets passed in doesn't currently have a trailing slash. (I'm not changing that behavior)

Take for example, on Windows
path is `config\cicd\base\01-namespaces\cicd-environment.yaml`

The `ToSlash` call will change this to
`config/cicd/base/01-namespaces/cicd-environment.yaml`
(on non-Windows platforms, it will be the same)

so the second argument to `TrimPrefix` must be 
`config/cicd/base/`     (with the trailing slash)

The result for `filename` will then be: `01-namespaces/cicd-environment.yaml`
and that is what we want.

Otherwise it will be `/01-namespaces/cicd-environment.yaml`

The kustomize file will have:

```
resources:
- 01-namespaces/cicd-environment.yaml
- 02-rolebindings/argocd-admin.yaml
- .....

```

wtam2018: Please comment the code.  It is not straight forward to see why ToSlash() is needed.

wtam2018: Please add a unit test as well.

keithchong: The unit test has been added.

PR #269, opened 2021-08-26
gitops-1169: Interactive mode with default option doesn't support push to git

ciiay:
**What type of PR is this?**

/kind enhancement

Fixes #

Day 1 operations both interactive mode and non-interactive mode are successful on OCP 4.9


chetan-rns:
```shell
? Select yes to accept default values or select no to be prompted for all options that haven't already been specified on the command line

Default values:
--gitops-repo-url               ""
--prefix			"p"
--service-repo-url		""
--save-token-keyring		"false"
--private-repo-driver		""
--gitops-webhook-secret		""
--output			"./gitops"
--git-host-access-token		""
--dockercfgjson			"~/.docker/config.json"
--image-repo			""
--overwrite			"false"
--service-webhook-secret	""
--push-to-git			"false"

```

Hi @ciiay, thanks for the PR. When users opt for smart defaults, kam will construct values for some of the above flags. For example, the `--image-repo ` will be constructed as `image-registry.openshift-image-registry.svc:5000/cicd/<service-repo>` but the default value will be displayed as " ". And users can also pass individual flags in the interactive mode which might change these values
`kam bootstrap --image-repo quay.io/cbanavik/taxi --interactive`. I wonder if there's a way we can handle this :thinking:

ciiay:
@chetan-rns thanks for the review and the feedback. Is it okay to hardcode the default values in the printout or should I get those values from somewhere in the code?

ciiay:
After a discussion with @keithchong and @chetan-rns , now the printout looks like this.
Screen Shot 2021-08-31 at 2 18 49 PM

pkg/cmd/bootstrap.go
```diff
@@ -39,6 +39,19 @@ const (
 	pipelinesOperatorName  = "OpenShift Pipelines Operator"
 )
 
+// Default flag values
+var defaultFlagValues = map[string]interface{}{
+	"gitops-webhook-secret":  "Auto-generated by kam if not user-overwritten", // Auto-generated by kam if not user-overwritten, no ""
+	"output":                 "\"./gitops\"",
+	"prefix":                 "\"\"",
+	"dockercfgjson":          "\"~/.docker/config.json\"",
+	"image-repo":             "Default values is internal registry", // Default values is internal registry, no ""
```
keithchong, line 48: Default value is

pkg/cmd/ui/ui.go
```diff
@@ -233,11 +235,22 @@ func SelectOptionPushToGit() bool {
 }
 
 // UseDefaultValues allows users to use default values so that they will be prompted with fewer questions in interactive mode
-func UseDefaultValues() bool {
+func UseDefaultValues(defaultFlagValues map[string]interface{}) bool {
+	// Print default values
+	flagValues := "\n\nDefault values (Can be overwritten from the command line):\n"
```
keithchong, line 240: Thought about this some more. Let's remove the parentheses:

The default values used for the options, if not overwritten from the command line, are:

pkg/cmd/bootstrap.go
keithchong: I think you can remove the comments now

pkg/cmd/bootstrap.go
```diff
@@ -39,6 +39,19 @@ const (
...
+	"overwrite":              "\"false\"",
+	"service-webhook-secret": "Auto-generated by kam if not user-overwritten", // Auto-generated by kam if not user-overwritten, no ""
```
keithchong, line 50: I think you can remove the comment now

ciiay:
Updated printout
Screen Shot 2021-09-02 at 11 43 19 AM

pkg/cmd/bootstrap.go
```diff
@@ -39,6 +39,19 @@ const (
 	pipelinesOperatorName  = "OpenShift Pipelines Operator"
 )
 
+// Default flag values
+var defaultFlagValues = map[string]interface{}{
```
chetan-rns, line 43: Since all the map values are of type string can we change the value type to string? 
```suggestion
var defaultFlagValues = map[string]string{
```

pkg/cmd/bootstrap.go
```diff
@@ -149,7 +162,7 @@ func missingFlagErr(flags []string) error {
 func initiateInteractiveMode(io *BootstrapParameters, client *utility.Client, cmd *cobra.Command) error {
 	log.Progressf("\nStarting interactive prompt\n")
 	// Prompt if user wants to use all default values and only be prompted with required or other necessary questions
-	promptForAll := !ui.UseDefaultValues()
+	promptForAll := !ui.UseDefaultValues(defaultFlagValues)
```
chetan-rns, line 165: Since we no longer use `defaultFlagValues` map in the bootstrap package, can we directly move it inside the function `UseDefaultValues()`?

chetan-rns (commented):
Everything else looks good @ciiay

chetan-rns (approved):
Thanks! @ciiay

/lgtm
/approve

PR #273, opened 2021-09-02
Scenarios for applying secrets and creating webhook

varshab1210:
**What type of PR is this?**

 /kind enhancement

**What does this PR do / why we need it**:

**Have you updated the necessary documentation?**

* [ ] Documentation update is required by this PR.
* [ ] Documentation has been updated.

**Which issue(s) this PR fixes**:

Fixes #?

**How to test changes / Special notes to the reviewer**:
`make e2e` should pass

PR #275, opened 2021-09-03
GITOPS-1000 Parse the image only if git repository is provided

ciiay:
**What type of PR is this?**

/kind bug

Added git repo url nil check for image parsing process.

Fixes #

`kam service add` command with no --git-repo-url flag is successful on OCP 4.8


chetan-rns:
@ciiay Running `make gofmt` should fix the format

chetan-rns (changes requested):
Thanks for the PR @ciiay. This will fix the issue. But after taking a closer look, kam is generating the webhook secret for the newly added service. I don't think the secret is being used anywhere since the service doesn't have a git repository. Should we consider adding a condition to avoid generating the webhook secret when the git repo is absent?

varshab1210:
In pr https://github.com/redhat-developer/kam/pull/273 I am observing the below error

```
Command stderr:  ✗  failed to bootstrap resources: failed to read Docker config "/var/run/kam-data/kam-test-secrets/*******-kambot-auth.json" : open /var/run/kam-data/kam-test-secrets/*******-kambot-auth.json: no such file or directory
```
I am retesting this pr for cross verification

/test all

chetan-rns (approved):
Thanks, @ciiay!
 
/lgtm
/approve

PR #277, opened 2021-09-15
Adding more scenarios for day1 operation

varshab1210:
**What type of PR is this?**

/kind enhancement

Added scenario for testing KAM bootstrap command with and without --overwrite flag. This pr also fixes the need to generate resource files with different names by kam bootstrap command (which are redundant) by cleaning up the directory before each scenario

Fixes #? N/A

```make e2e``` should pass

### DIFF
--- a/test/e2e/kamsuite/kamsuite.go
+++ b/test/e2e/kamsuite/kamsuite.go
@@ -64,6 +64,14 @@ func FeatureContext(s *godog.Suite) {
 
 	s.AfterScenario(func(*messages.Pickle, error) {
 		fmt.Println("After scenario")
+		err := os.RemoveAll("bootstrapresources")
+		if err != nil {
+			log.Fatal(err)
+		}
+		err2 := os.RemoveAll("secrets")
+		if err2 != nil {
+			log.Fatal(err)
+		}
 		re := regexp.MustCompile(`[a-z]+`)
 		scm := re.FindAllString(os.Getenv("GITOPS_REPO_URL"), 2)[1]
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: test/e2e/kamsuite/kamsuite.go [hunk 1]
```diff
@@ -139,6 +144,24 @@ func deleteGitlabRepository(arg []string) (bool, string) {
 	return true, stderr.String()
 }
 
+func ClearDir() error {
```
chetan-rns, line 147:
```suggestion
func clearDir() error {
```
The function is exported but only used within the same package

test/e2e/kamsuite/kamsuite.go
```diff
@@ -64,6 +65,10 @@ func FeatureContext(s *godog.Suite) {
 
 	s.AfterScenario(func(*messages.Pickle, error) {
 		fmt.Println("After scenario")
+		err := ClearDir()
```
chetan-rns, line 68: If we are trying to delete the gitops folder `os.RemoveAll(path)`  will remove the entire folder. WDYT?

chetan-rns (changes requested):

test/e2e/kamsuite/kamsuite.go
varshab1210: I had tried it that way but it would remove even ```test-run``` folder and cause test failure.

varshab1210: Thanks for the advice @chetan-rns , will do

chetan-rns (approved):
Thanks @varshab1210 !

/lgtm
/approve

PR #295, opened 2022-01-21
fix: sanitize trigger binding name

chetan-rns:
**What type of PR is this?**

> /kind bug

Sanitize the trigger binding name so it's a valid DNS1035 name

Fixes #?

1. Add an env and a service with long names
2. Create Webhook cmd should suceed


ishitasequeira (approved):
Thank for the PR @chetan-rns!!
LGTM

PR #298, opened 2022-03-03
chore: add flags for external linker to satisfy RPMdiff bind-now test

jaideepr97:
**What type of PR is this?**

> /kind bug
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

it seems like kam pulls in the `net` package as a dependency which I suspect is causing `cgo` to be used for compilation indirectly and maybe causing kam to be dynamically linked. (This might also be putting the KAM binary in a position to benefit from supplying the `-buildmode=pie` flag) The [RPMDiff test](https://rpmdiff.engineering.redhat.com/run/527677/7/) in cpaas currently throws an error because we don't add the `-Wl,-z,now` flags for external linker while building the kam binary 
This PR fixes that 

Fixes https://issues.redhat.com/browse/GITOPS-1714 partially

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ DARWIN=$(EXECUTABLE)_darwin_amd64
 PKGS := $(shell go list  ./... | grep -v test/e2e | grep -v vendor)
 FMTPKGS := $(shell go list  ./... | grep -v vendor)
 VERSION=$(shell git describe --tags --always --long --dirty)
-LD_FLAGS="-s -w -X github.com/redhat-developer/kam/pkg/cmd/version.Version=$(VERSION)"
+LD_FLAGS='-s -w -X github.com/redhat-developer/kam/pkg/cmd/version.Version=$(VERSION) -extldflags "-Wl,-z,now"'
 
 .PHONY: all_platforms
 all_platforms: windows linux darwin 


jannfis:
I was wondering tho, if we could just not set `CGO_ENABLED=0` for the build?

jaideepr97:
@jannfis Would that be a problem if we are using any packages that absolutely require `cgo`? My understanding is the `net` package that is pulled in by one of kam's dependencies maybe one such package but I don't have an in-depth understanding of this

jaideepr97:
I was thinking we could just try to build a new RPM with these extdflags in place, supply the `buildmode=pie` flag and then find a way to run the RPMDiff test again just on this newly built RPM to see if that addresses all the concerns

jannfis:
> Would that be a problem if we are using any packages that absolutely require cgo

Yes, probably

> My understanding is the net package that is pulled in by one of kam's dependencies maybe one such package

I think it used to be due to the resolver, but meanwhile, Go also provides a Go-native resolver to be used with CGO_ENABLED=0 (https://pkg.go.dev/net#hdr-Name_Resolution)

kam compiles, links and runs fine when built with `CGO_ENABLED=0`. However, it produces a static binary. What's funny is, that it's even a little smaller than the dynamically one, which was built using cgo:

```
-rwxr-xr-x 1 foo foo 45056000 Mar  3 21:54 dist/kam_linux_amd64
-rwxr-xr-x 1 foo foo 45109248 Mar  3 21:53 dist/kam_linux_amd64-cgo
```

Anyway, I think for now, we should go for the linker flags.

PR #310, opened 2022-04-26
Build for linux/ppc64le and linux/s390x

yselkowitz:
**What type of PR is this?**

/kind enhancement

Enables builds of kam for linux/ppc64le and linux/s390x.


Makefile
```diff
@@ -17,16 +17,17 @@ all_platforms: windows linux darwin
...
 .PHONY: darwin
 darwin: $(DARWIN)
 
 $(WINDOWS):
 	env GOOS=windows GOARCH=amd64 go build -o $(DIST_DIR)/$(WINDOWS)  -ldflags=$(LD_FLAGS)  cmd/kam/kam.go
 
-$(LINUX):
-	env GOOS=linux GOARCH=$(ARCHITECTURE) go build -o $(DIST_DIR)/$(LINUX) -ldflags=$(LD_FLAGS) cmd/kam/kam.go
+linux:
+	env GOOS=linux GOARCH=amd64 go build -o $(DIST_DIR)/$(EXECUTABLE)_linux_amd64 -ldflags=$(LD_FLAGS) cmd/kam/kam.go
```
wtam2018, line 28: I would stick with the pattern.  So, it looks something like this.

```
$(LINUX): 
    env GOOS=linux GOARCH=$(ARCHITECTURE) go build -o $(DIST_DIR)/$(EXECUTABLE)_linux_$(ARCHITECTURE)  -ldflags=$(LD_FLAGS) cmd/kam/kam.go

yselkowitz: The point is that we now want to build multiple linux architectures simultaneously, not just one, so the existing pattern needs to change one way or another.  There are various ways this could look.  For instance, if you still want individual targets for each supported os/arch  combo, that could be done instead, but other things will change as a result.

Keep in mind as well that darwin/arm64 likely will also need to be added at some point soon for the new ARM Macs, which will then disrupt the pattern there too.

wtam2018: Agreed.

PR #323, opened 2022-07-21
add documentation about DCO

reginapizza:
**What type of PR is this?**
> /kind documentation

Now that DCO checks are enabled, this will add documentation about it as requested by Red Hat Legal team. 

https://issues.redhat.com/browse/GITOPS-2122


wtam2018 (commented):
We need to come up with our own Contributing.md. The Linux one does not really apply 100%.

reginapizza:
@wtam2018 Red Hat Legal states the following in their [FAQ](https://docs.google.com/document/d/1_PlK0GfAuKBk6ruysTo57qQGbCRzjqRWMiW50mqwdfg/edit): 

> 11. Can we modify or customize the DCO certification text?
No (unless you get approval from Red Hat Legal). The Linux kernel project has requested that modified DCOs not be used so that the community can standardize on one text.

wtam2018:
I thought you were granting copyright our software to
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129
when I first looked at it.  It was actually the copyright of the DCO.  I think you can just use this link. https://developercertificate.org/

reginapizza:
haha no worries, that wouldn't be good! Also in the FAQ they state: 
> While there are various ways of doing this, we currently recommend the following approach. 
> First, add a file to the root directory of your project repository containing just the text of version 1.1 of the DCO (example: https://github.com/metal3-io/metal3-docs/blob/master/DCO).

I took the text here from the example given above. The only different between the text at https://developercertificate.org/ and here is the additional address. If you really don't think it should be there, let me know and I'll change it but otherwise I think it's a minor difference.

CONTRIBUTING.md
```diff
@@ -0,0 +1,6 @@
+## Certificate of Origin
+
+By contributing to this project you agree to the Developer Certificate of
+Origin (DCO). This document was created by the Linux Kernel community and is a
+simple statement that you, as a contributor, have the legal right to make the
+contribution. See the [DCO](DCO) file for details.
```
wtam2018, line 6:
```suggestion
contribution. See the [DCO](https://developercertificate.org/) file for details.
```

DCO
```diff
@@ -0,0 +1,37 @@
+Developer Certificate of Origin
```
wtam2018, line 1: And we don't need this md file

PR #327, opened 2022-11-30
Build Linux, MacOS, and Windows ARM 64 binaries

yselkowitz:
**What type of PR is this?**

/kind enhancement

This is the first step in enabling ARM kam binaries.  Another PR to release and serve the binaries will follow.

N/A, not user-visible yet.

`make all_platforms` should succeed, and create linux/arm64, darwin/arm64, and windows/arm64 binaries.


wtam2018:
@yselkowitz the change looks good. Any idea which source file failed the Format Check?

yselkowitz:
> @yselkowitz the change looks good. Any idea which source file failed the Format Check?

It only occurs with golang 1.19; filed: https://github.com/redhat-developer/kam/pull/328

Issue #344, opened 2023-09-08
E2E test failures with latest version of Openshift-Pipelines v1.11^


svghadi:
**Describe the bug**
The kam e2e test suite is currently failing due to failure to sync the deployed `cicd-app` application. The tests use `.spec.resources` in Tekton tasks & pipelines. However, the latest version of Openshift-Pipelines (v1.11^) has [dropped](https://docs.openshift.com/container-platform/4.13/cicd/pipelines/op-release-notes.html#deprecated-features-1-11_op-release-notes) support for this field. 

We need to update our tekton tasks & pipelines stored in [kam-bot/bus](https://github.com/kam-bot/bus) to use replacement tasks. 
- https://tekton.dev/docs/pipelines/pipelineresources/#replacing-pipelineresources-with-tasks

**To Reproduce**
Follow [test guide](https://github.com/redhat-developer/kam/blob/master/docs/test/test-guide.md) to run e2e.

**Screenshots**

**Workaround**
A workaround to unblock the e2e tests and CI is to use Openshift-Pipelines v1.10 for testing. However, this version is supported only up to Openshift 4.13. Therefore, testing on Openshift 4.14^ will still result in failures.

PR #345, opened 2023-09-08
Pin openshift-pipelines version to 1.10 for testing on OCP 4.12 & 4.13

svghadi:
**What type of PR is this?**
 /kind bug

Related #344

In latest versions of openshift-pipelines, some fields are dropped which causes kam test failures. This PR adds a workaround to unblock CI by pinning openshift-pipelines version to 1.10 for testing on OCP 4.12 & 4.13.


iam-veeramalla (approved):
/lgtm
/approve

PR #346, opened 2023-10-12
Do away with HTTP server and instead used a Go based File server for hosting binaries

anandf:
**What type of PR is this?**
/kind enhancement

The kam container image uses `httpd` base image which requires frequent updates for CVE fixes and is difficult to maintain. By using a Go based fileserver for hosting the binaries for different OS and architecture, we can do away with HTTP server and instead use a light weight image. Even a `scratch` base image should be enough  for running a go binary.

Fixes #?


wtam2018 (commented):
/lgtm
/approve

